PR #35, opened 2022-05-04
34 feature request support raw queries

### LINKED ISSUE #34: feature request: support raw queries
I would like another quantifier for [raw queries](https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Client.html#method.query_raw). It allow me to perform calculation whiteout allocating a Vec or mapping rows before allocating.

LouisGariepy:
Closes #34.

The implementation is **not yet complete**, and as of now **incorrect** with respect to the changes proposed here
- [x] https://github.com/LouisGariepy/cornucopia/issues/34#issuecomment-1117381238 (returning a stream with strongly typed items, love it). 
- [x] https://github.com/LouisGariepy/cornucopia/issues/34#issuecomment-1117386309 (Improving queries returning vec to use `query_raw` to avoid unnecessary allocations).
- [x] I'll need to make sure the grammar also works with these changes, as I have assumed the "raw" identifier would be a "return_type", but now that we can strongly type the return stream, it would probably make a lot more sense to use it as a quantifier (as was originally proposed :) )

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,7 +161,7 @@ version = "0.6.1"
 dependencies = [
  "async-recursion",
  "clap",
- "cornucopia_client 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cornucopia_client 0.1.0",
  "deadpool-postgres",
  "eui48",
  "heck",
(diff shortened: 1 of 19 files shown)


LouisGariepy:
Added some tests, everything seems to be working fine. Implementation is a bit messy, but we're approaching a state where the crate might become stable soon, some I'm thinking of cleaning up everything at that point. In any case, the generated code is pretty neat, which is what matters the most

Issue #48, opened 2022-05-12
bug: binary operation `==` cannot be applied to type `Vec<_>` where `_` was generated by cornucopia


xoac:
After generate code I am getting this error:
```
error[E0369]: binary operation `==` cannot be applied to type `Vec<public::Transfer>`
   --> geyser-plugin/src/cornucopia.rs:289:13
    |
285 |         #[derive(Debug, Clone, PartialEq)]
    |                                --------- in this derive macro expansion
...
289 |             pub transfers: Vec<super::super::types::public::Transfer>,
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this error originates in the derive macro `PartialEq` (in Nightly builds, run with -Z macro-backtrace for more info)
````
generated inside `types` module:
```rust
#[derive(Debug, postgres_types::ToSql, postgres_types::FromSql)]
#[postgres(name = "transfer")]
#[derive(Clone)]
pub struct Transfer {
    pub mint: Vec<u8>,
    pub to_account: Vec<u8>,
    pub from_account: Vec<u8>,
    pub amount: i64,
}
```

generated inside `queries` module:
```rust
#[derive(Debug, Clone, PartialEq)]
pub struct GetTransactionCacheByAnySigner {
    pub signature: Vec<u8>,
    pub fee: i64,
    pub transfers: Vec<super::super::types::public::Transfer>,
    pub signers: Vec<Vec<u8>>,
    pub status: super::super::types::public::TransactionStatus,
    pub slot: i64,
    pub updated_on: time::PrimitiveDateTime,
}
```

The solution to this is to add `PartialEq` to `Transfer`

LouisGariepy:
Hi @xoac, thanks for opening this issue :smile: 

I'm currently working on a complete rearchitecture of Cornucopia, so I'll add this to my list of things to do before 0.8 :+1: 

In general, regarding adding more `derive`s, it's not necessarily trivial to know which trait can be derived for which struct. That is, I have to make sure either the derive works for every single type that cornucopia supports, or create a mechanism to determine which bounds can be added.

Either way, its not a blocker, but I'll have to pour some more thoughts into this. It will be done though soonish though, that I can say.

LouisGariepy:
@xoac This should be fixed in the next release that should come out in a few days :tada:

LouisGariepy:
On track to be solved by #52.

Issue #47, opened 2022-05-11
Struct parameters


LouisGariepy:
### Discussed in https://github.com/LouisGariepy/cornucopia/discussions/44

<div type='discussions-op-text'>

<sup>Originally posted by **Virgiel** May 11, 2022</sup>
In one of my projects, I have long queries like this one:
```sql
INSERT INTO pro 
  (title, fname, lname, sex, photo, specialty, expertise, street, postcode, city, phone, link, lat, lng, source, show, trait)
  VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17) 
  RETURNING id;
```

And I need code like this one to pass all arguments: 
```rs
[
    &item.pro.title.as_str(),
    &item.pro.fname.as_str(),
    &item.pro.lname.as_str(),
    &item.pro.sex,
    &item.pro.photo.as_ref().map(Photo::name),
    &item.pro.specialty.as_ref(),
    &item.pro.expertise.as_ref(),
    &item.pro.street.as_str(),
    &item.pro.postcode.as_str(),
    &item.pro.city.as_str(),
    &item.pro.phone.as_ref().map(ToString::to_string),
    &item.pro.link.as_ref().map(ToString::to_string),
    &item.pro.lat,
    &item.pro.lng,
    &item.source.as_ref(),
    &item.show,
    &item.pro.r#trait.as_ref(),
]
```

In addition to being cumbersome, it is also error-prone, every time I change the query I can easily swap any column of the same time without any compiler error. However, I like how small and succinct cornucopia functions can be and we should keep that.

My idea is the following. This code :
```sql
--! example (hello, world) #
/* some fancy SQL */
```

Generate this:
```rs
fn example(hello: &str, world: &i32) -> /* Query or Result */ {
    /* some fancy Rust */
}

struct ExampleParams<'a> {
    hello: &'a str,
    world: &'a i32
}

impl <'a> ExampleParams<'a> {
    fn example(self) -> /* Query or Result */ {
        example(self.hello, self.world)
    }
}
```

We could do even crazier things by allowing user naming of params struct.

```sql
--! insert (TestParams) #
/* some fancy SQL */

--! update (TestParams) #
/* some fancy SQL */
```

```rs
fn insert(hello: &str, world: &i32) -> /* Query or Result */ {
    /* some fancy Rust */
}

fn update(world: &i32, hello: &str) -> /* Query or Result */ {
[...]

Issue #46, opened 2022-05-11
No quantifier design


LouisGariepy:
### Discussed in https://github.com/LouisGariepy/cornucopia/discussions/42

<div type='discussions-op-text'>

<sup>Originally posted by **Virgiel** May 10, 2022</sup>
Currently, every query has to declare a quantifier but we could do without it. The idea is for queries to return a query struct that defer execution.

This code:
```sql
--! authors()
SELECT * FROM Author;
```

Generate this function:
```rs
pub fn authors<'a, C: GenericClient>(client: &'a C) -> Query<'a, C, (i32, String, String)> {
    Query {
        query: "SELECT * FROM Author;",
        client,
        params: &[],
        mapper: |res| {
            let return_value_0: i32 = res.get(0);
            let return_value_1: String = res.get(1);
            let return_value_2: String = res.get(2);
            (return_value_0, return_value_1, return_value_2)
        },
    }
}
```

With this client code:
```rs

pub struct Query<'a, C: GenericClient, T> {
    query: &'static str,
    client: &'a C,
    params: &'a [&'a (dyn tokio_postgres::types::ToSql + Sync)],
    mapper: fn(Row) -> T,
}

impl<'a, C, T: 'static> Query<'a, C, T>
where
    C: GenericClient,
{
    pub async fn stmt(&self) -> Result<Statement, Error> {
        self.client.prepare(self.query).await
    }

    pub async fn one(self) -> Result<T, Error> {
        let stmt = self.stmt().await?;
        let row = self.client.query_one(&stmt, self.params).await?;
        Ok((self.mapper)(row))
    }

    pub async fn opt(self) -> Result<Option<T>, Error> {
        let stmt = self.stmt().await?;
        Ok(self
            .client
            .query_opt(&stmt, self.params)
            .await?
            .map(self.mapper))
    }

    pub async fn list(self) -> Result<Vec<T>, Error> {
        self.raw().await?.try_collect().await
    }

    pub async fn raw(self) -> Result<impl Stream<Item = Result<T, Error>>, Error> {
        let stmt = self.stmt().await?;
        let stream = self
            .client
            .query_raw(&stmt, slice_iter(self.params))
            .await?
            .map(move |res| res.map(self.mapper));
        Ok(stream.into_stream())
    }
}

// https://github.com/sfackler/rust-postgres/blob/765395f288861209a644c621bf72172acd482515/tokio-postgres/src/lib.rs
fn slice_iter<'a>(
[...]

Issue #45, opened 2022-05-11
Named parameters


LouisGariepy:
### Discussed in https://github.com/LouisGariepy/cornucopia/discussions/43

<div type='discussions-op-text'>

<sup>Originally posted by **Virgiel** May 11, 2022</sup>
In one of my projects, I have long queries like this one:
```sql
INSERT INTO pro 
  (title, fname, lname, sex, photo, specialty, expertise, street, postcode, city, phone, link, lat, lng, source, show, trait)
  VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17) 
  RETURNING id;
```

Having to repeat all the names is cumbersome, even more, when I need to do it for all inserts and updates. I found your previous Reddit [comment](https://www.reddit.com/r/rust/comments/u64cy5/comment/i58qp2g/?utm_source=share&utm_medium=web2x&context=3) on that matter. I understand the appeal of using valid SQL, but i think we should also support named parameters.

We could use MySQL syntax `@name` that should be highlighted by code editor, an extended Postgres one `$name` or another syntax found in SQLDelight `:name`. Indexed and named parameters should be incompatible, so you can either write
```sql
--! insert_pro(title, fname, ..., trait)
INSERT INTO pro 
  (title, fname, ..., trait)
  VALUES ($1, $2, ..., $17) 
  RETURNING id;
```

or 
```sql
--! insert_pro()
INSERT INTO pro 
  (title, fname, ..., trait)
  VALUES (:title, :fname, ..., :trait) 
  RETURNING id;
```
</div>

LouisGariepy:
On track to be solved by #52.

PR #52, opened 2022-05-17
New query architecture

### LINKED ISSUE #25: Provide a way for users to cutomize `derive`s on custom types
I don't want to presuppose too much about how the generated code will be used, thus, so far, I've refrained from adding any superfluous derives to the generated types.

This might be limiting to some users wishing to add `derive`s to their types (like `Serialize` or `Clone`). I'm thinking of flags like `--custom-struct-derive`, `--custom-enum-derive` and `--custom-derive` which could be used to add the `derive` only to `struct`s, `enum`s, or both.

This will require a bit more thought though, so if you're reading this thread because you want this feature, feel free to ping me and explain how you'd like this feature to work in practice.

LouisGariepy:
This PR tracks the progress the new query architecture for the upcoming v0.8. This version is a significant re-architecture so there are quite a few things to do to ensure a smooth release. Some of the relevant discussions here #45, #46, and #47.

- [x] Overhaul parser
- [x] Overhaul type registration
- [x] Implement query generation
- [x] Update client code
- [x] Comment/Improve source (not perfect but its barely ok)
- [x] Update examples
- [x] Update tests
- [x] Update README

Closes #25 #39 #40 #41 #45 #46 #47 #48 #56


LouisGariepy:
Composite and domain custom types don't play well with the `ToSql` and `FromSql` macro derives. I'll have to generate an impl instead of relying on the macros. I'll also need to generate an owned and borrowed version of custom types (only owned if the type is Copy).

On it! :smile_cat:

LouisGariepy:
Also, I need to make sure arrays coming from postgres are never considered as borrowable (because `tokio_postgres` will deserialize arrays either as Vec or [T;N], not slices or iterators). This has nothing to do with Cornucopia per se, just a perfomance limitation of deserializing PostgreSQL arrays, so overall not a big deal.

Virgiel:
I never thought about arrays :/ Unfortunately, the size of arrays in the schema is only informative but not [enforced](https://www.postgresql.org/docs/current/arrays.html). We should not use [T; N].

Virgiel:
What if I told you we can have an iterator for arrays ?

Let start by stealing some code from `postgres-type` because they love to have everything private:

```rs
/// https://github.com/sfackler/rust-postgres/blob/38da7fa8fe0067f47b60c147ccdaa214ab5f5211/postgres-protocol/src/types/mod.rs

/// An iterator over the dimensions of an array.
pub struct ArrayDimensions<'a>(&'a [u8]);

impl<'a> FallibleIterator for ArrayDimensions<'a> {
    type Item = ArrayDimension;
    type Error = Box<dyn std::error::Error + Sync + Send>;

    #[inline]
    fn next(&mut self) -> Result<Option<ArrayDimension>, Box<dyn std::error::Error + Sync + Send>> {
        if self.0.is_empty() {
            return Ok(None);
        }

        let len = self.0.read_i32::<BigEndian>()?;
        let lower_bound = self.0.read_i32::<BigEndian>()?;

        Ok(Some(ArrayDimension { len, lower_bound }))
    }

    #[inline]
    fn size_hint(&self) -> (usize, Option<usize>) {
        let len = self.0.len() / 8;
        (len, Some(len))
    }
}

/// Information about a dimension of an array.
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
pub struct ArrayDimension {
    /// The length of this dimension.
    pub len: i32,

    /// The base value used to index into this dimension.
    pub lower_bound: i32,
}

/// An iterator over the values of an array, in row-major order.
pub struct ArrayValues<'a> {
    remaining: i32,
    buf: &'a [u8],
}

impl<'a> FallibleIterator for ArrayValues<'a> {
    type Item = Option<&'a [u8]>;
    type Error = Box<dyn std::error::Error + Sync + Send>;

    #[inline]
    fn next(
        &mut self,
    ) -> Result<Option<Option<&'a [u8]>>, Box<dyn std::error::Error + Sync + Send>> {
        if self.remaining == 0 {
            if !self.buf.is_empty() {
                return Err("invalid message length: arrayvalue not drained".into());
            }
            return Ok(None);
        }
        self.remaining -= 1;

        let len = self.buf.read_i32::<BigEndian>()?;
        let val = if len < 0 {
            None
        } else {
            if self.buf.len() < len as usize {
                return Err("invalid value length".into());
            }

            let (val, buf) = self.buf.split_at(len as usize);
            self.buf = buf;
            Some(val)
        };

        Ok(Some(val))
    }

    fn size_hint(&self) -> (usize, Option<usize>) {
        let len = self.remaining as usize;
        (len, Some(len))
    }
}

/// A Postgres array.
pub struct Array<'a> {
    dimensions: i32,
    has_nulls: bool,
    element_type: Oid,
    elements: i32,
    buf: &'a [u8],
}

impl<'a> Array<'a> {
    /// Returns true if there are `NULL` elements.
    #[inline]
    pub fn has_nulls(&self) -> bool {
        self.has_nulls
    }

    /// Returns the OID of the elements of the array.
    #[inline]
    pub fn element_type(&self) -> Oid {
        self.element_type
    }

    /// Returns an iterator over the dimensions of the array.
    #[inline]
    pub fn dimensions(&self) -> ArrayDimensions<'a> {
        ArrayDimensions(&self.buf[..self.dimensions as usize * 8])
    }

    /// Returns an iterator over the values of the array.
    #[inline]
    pub fn values(&self) -> ArrayValues<'a> {
        ArrayValues {
            remaining: self.elements,
            buf: &self.buf[self.dimensions as usize * 8..],
        }
    }
}
```

Now we can make our own array iterator type:
```rs
struct ArrayIterator<'a, T: FromSql<'a>> {
    values: ArrayValues<'a>,
    ty: Type,
    _type: PhantomData<T>,
}

impl<'a, T: FromSql<'a>> Iterator for ArrayIterator<'a, T> {
    type Item = T;

    fn next(&mut self) -> Option<Self::Item> {
        self.values
            .next()
            .unwrap()
            .map(|raw| T::from_sql_nullable(&self.ty, raw).unwrap())
    }
}

impl<'a, T: FromSql<'a>> FromSql<'a> for ArrayIterator<'a, T> {
    fn from_sql(
        ty: &Type,
        raw: &'a [u8],
    ) -> Result<ArrayIterator<'a, T>, Box<dyn std::error::Error + Sync + Send>> {
        let member_type = match *ty.kind() {
            Kind::Array(ref member) => member,
            _ => panic!("expected array type"),
        };

        let array = array_from_sql(raw)?;
        if array.dimensions().count()? > 1 {
            return Err("array contains too many dimensions".into());
        }

        Ok(ArrayIterator {
            ty: member_type.to_owned(),
            values: array.values(),
            _type: PhantomData::default(),
        })
    }
    
    fn accepts(ty: &postgres_types::Type) -> bool {
        match *ty.kind() {
            Kind::Array(ref inner) => T::accepts(inner),
            _ => false,
        }
    }
}
```

And taaadam:
```rs
let all = client
    .query_raw("SELECT * FROM test", std::iter::empty::<i32>())
    .await
    .unwrap();
pin_mut!(all);
while let Some(row) = all.try_next().await.unwrap() {
    let foo: ArrayIterator<&str> = row.get(0);
    let vec: Vec<_> = foo.collect();
    println!("foo: {:?}", vec);
}
```

Virgiel:
This is just a draft (lots of unwrap and questionable to_owned) and it's a shame to have to steal so much code but how about doing something like this?

LouisGariepy:
Wow! Thank you! I have to read the code more carefully, but it'd be a major improvement :astonished: Gonna try it out.

In a similar vein, it seems `&serde_json::Value` (notice the borrow) does not implement `FromSql`. Not sure if there's a good reason or if it's just oversight, but we might be able to solve it.

Virgiel:
Internally `serde_json::Value` own [data](https://docs.rs/serde_json/latest/serde_json/enum.Value.html) :/
I think it is preferable for you to finish a version with always owned arrays and json, so we can make an alpha version, and then we will try to reduce allocation later until release.

Virgiel:
We could use our custom borrowing `FromSql` and try to upstream them into `postgres`.

LouisGariepy:
Apart from the unwraps (didn't check them carefully yet), this looks good!

> to_owned()

Don't think we can do without, but surely taking ownership of one `member_type` is better than allocating the whole array.

> Internally serde_json::Value own [data](https://docs.rs/serde_json/latest/serde_json/enum.Value.html) :/

Yeah, that's too bad :/

> I think it is preferable for you to finish a version with always owned arrays and json, so we can make an alpha version, and then we will try to reduce allocation later until release.

I'll give a single shot at your Array implementation, but I won't delay any further than that. Then I'll publish ASAP and then keep working on tests, docs, etc.

Thanks again for your contributions, as mentioned in #57 I'll soon be able to accept contributors in an official capacity.

LouisGariepy:
> We could use our custom borrowing FromSql and try to upstream them into postgres.

Good plan.

Virgiel:
My iterator solution has a big problem. We need a failing iterator that handles errors instead of using unwrap but the error returned is a boxed std error and we would need a `tokio-postgres::Error` to be orthogonal with the library API. But `tokio-postgres::Error` is private :/

So many things would be easier if more things were public in `postgres`, like `tokio-postgres::Error`, `slice_iter` and all the Array logic. We could ask the maintainer because I think we have a compelling use case.

Virgiel:
One solution could be to temporarily [override](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) `tokio-postgres` with patches. I don't want to create a real fork, but just relax some internal privacy for our use.

LouisGariepy:
Sorry I'm not totally following. Are we talking about these unwraps?

```rust
impl<'a, T: FromSql<'a>> Iterator for ArrayIterator<'a, T> {
    type Item = T;

    fn next(&mut self) -> Option<Self::Item> {
        self.values
            .next()
            .unwrap() //<-- this
            .map(|raw| T::from_sql_nullable(&self.ty, raw).unwrap()) <-- that
    }
}
```

Virgiel:
Both

LouisGariepy:
@Virgiel Yeah that's what I was assuming too. If this is the case, it fine by me to panic/unwrap.

Virgiel:
> In a similar vein, it seems `&serde_json::Value` (notice the borrow) does not implement `FromSql`. Not sure if there's a good reason or if it's just oversight, but we might be able to solve it.

What about [RawValue](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html) ? Should do the trick !

LouisGariepy:
> > In a similar vein, it seems `&serde_json::Value` (notice the borrow) does not implement `FromSql`. Not sure if there's a good reason or if it's just oversight, but we might be able to solve it.
> 
> What about [RawValue](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html) ? Should do the trick !

Something like this? 
```rust

struct JsonBorrowed<'a>(&'a RawValue);

impl<'a> FromSql<'a> for JsonBorrowed<'a> {
    fn from_sql(
        ty: &Type,
        mut raw: &'a [u8],
    ) -> Result<JsonBorrowed<'a>, Box<dyn std::error::Error + Sync + Send>> {
        if *ty == Type::JSONB {
            let mut b = [0; 1];
            raw.read_exact(&mut b)?;
            // We only support version 1 of the jsonb binary format
            if b[0] != 1 {
                return Err("unsupported JSONB encoding version".into());
            }
        }
        serde_json::de::from_slice(raw)
            .map(JsonBorrowed)
            .map_err(Into::into)
    }

    accepts!(JSON, JSONB);
}

```

LouisGariepy:
> Erratum
> Array logic is public in postgres-protocol

:open_mouth:

Virgiel:
> Something like this?

There is already a [type](https://docs.rs/tokio-postgres/latest/tokio_postgres/types/struct.Json.html) for `Deserialize` from `tokio-prostgres`:
```rs
 let json: tokio_postgres::types::Json<&RawValue> = row.get(0);
```

Virgiel:
> Something like this?

There is already a [type](https://docs.rs/tokio-postgres/latest/tokio_postgres/types/struct.Json.html) for `Deserialize` from `tokio-postgres`:
```rs
 let json: tokio_postgres::types::Json<&RawValue> = row.get(0);
```

LouisGariepy:
Alright, the code generate works, at least to generate queries for the code we already had.

I have to stress that the code is pretty unsightly (understatement), but it is there and it does seem to work. Tomorrow I'll be able to clean up the code and check the other items for this PR. Cheers!

LouisGariepy:
(Tests are disabled right now, which is why the CI marks them as passing)

LouisGariepy:
Still a bit of an issue with borrowing wrapper types, but I know how to solve it I think. A lot of progress has been made though.

LouisGariepy:
I think codegen is good to go now (albeit, as I said above, in a very unruly form). Will try to wire some tests tonight...

Virgiel:
Sorry for the last minute proposal today, I hope I didn't waste too much of your time. Thanks for your hard work on this rewrite!

LouisGariepy:
There's this weird interaction in `tokio_postgres` which results in PostgreSql domains sometimes being interpreted as the proper Rust newtype, but also sometimes as the inner type. This seems to happen only in return columns, but I'm not sure. It's not a bug either, the behaviour is fine, it's just a little unexpected.

In any case this is pretty niche and harmless, and it's not something we can fix within cornucopia. Eventually, I could try to reach out to `tokio_postgres` to better understand what's up.

For now, I won't block further on this since I don't consider it to be a bug within Cornucopia.

LouisGariepy:
I messed up the implementation of `FromSql` for borrowed custom types, but that should be pretty easy to fix, and appart from that, all tests pass :partying_face:

Virgiel:
> There's this weird interaction in `tokio_postgres` which results in PostgreSql domains sometimes being interpreted as the proper Rust newtype, but also sometimes as the inner type. This seems to happen only in return columns, parameters are inferred correctly. It doesn't cause errors within Cornucopia, the behaviour is fine, it's just a little unexpected. Worst case scenario the user has to map this one field to wrap it in their newtype. Not ideal, but in the end it is pretty niche and harmless.

I couldn't find an explicit specification in the PostgreSQL documentation, but it seems from the questions and answers on the Internet that this is the correct behavior.
The [documentation](https://www.postgresql.org/docs/current/domains.html) states that "When an operator or function of the underlying type is applied to a domain value, the domain is automatically down-cast to the underlying type." This may be what happens on return. 
 So no problem with that

Virgiel:
After you push these changes I will be interested in writing a more complex test suite, especially one that tests we detect errors and show a well-formatted error message

LouisGariepy:
> I messed up the implementation of `FromSql` for borrowed custom types

Fixed!

LouisGariepy:
I just realized I forgot to handle nullable columns. Should be quick.

LouisGariepy:
Now passes CI :rocket:

LouisGariepy:
Unless I missed something, the only remaining tasks are doc-related.

Virgiel:
While trying to torture the codegen I found that nullable and array do not work well together:

```sql
-- Write your migration SQL here
CREATE TYPE nightmare_composite AS (
    custom custom_composite[],
    spongebob spongebob_character[]
);

CREATE TABLE nightmare (
    composite nightmare_composite NOT NULL,
    name TEXT NOT NULL,
    names TEXT[] NOT NULL,
    data BYTEA,
    datas BYTEA[]
);

INSERT INTO nightmare (composite, name, names)
    VALUES (
        ROW(
            CAST (ARRAY[ROW('incredible', 42, 'Patrick')] as custom_composite[]),
            CAST (ARRAY['Bob', 'Patrick'] as spongebob_character[])
            ), 
        'Bob', 
        ARRAY['I', 'Love', 'Chocolate']
    );
```

```sql
--! nightmare ?{data, datas}
SELECT * FROM nightmare;
```

Virgiel:
We also need a way to specify if items in arrays or columns in composite types are nullable, but this is not blocking this PR.

Virgiel:
If codegen is no longer having nightmares, I think it's ready for a merge

Issue #55, opened 2022-05-18
Expose a library equivalent to the CLI.


LouisGariepy:
Suggested in this discussion https://github.com/LouisGariepy/cornucopia/discussions/53.

This would allow Cornucopia to be used more ergonomically in automated environments, build scripts, and even other rust crates.

Virgiel:
The problem is that we need a working Postgres database. I found [pg-embed](https://github.com/faokunega/pg-embed), but I don't like the idea of downloading and running a binary from an untrusted source.

LouisGariepy:
My intention with this issue was to allow users to do something like :
```rust
let cornucopia = cornucopia::Cornucopia;
let generated_code = cornucopia
    .url("some/cool/url")
    //.podman() <-- or this
    //.docker() <-- or that
    .query_dir("...")
    .generate();
```
That would work without an embedded DB. I probably misunderstood what feature you really wanted though (sorry :smile:), so please let me know!

Virgiel:
This would be useful, especially when using build scripts! I was also thinking of a built-in option, but it's hard to do this properly with Postgres and can be done later.

PR #61, opened 2022-05-22
Test suite

### LINKED ISSUE #60: Error test suite
### Discussed in https://github.com/LouisGariepy/cornucopia/discussions/59

<div type='discussions-op-text'>

<sup>Originally posted by **Virgiel** May 22, 2022</sup>
It might be interesting to have a test suite that checks that we still catch errors after refactoring. I've started experimenting with this in my [fork](https://github.com/Virgiel/cornucopia/tree/test_suite).

My idea is to write some Toml files describing inputs and the expected error:
```toml
[[test]]
name = "PgParamInExtendedQuery"
query = """
--! new_author
INSERT INTO Author (id, name) VALUES (:id, $1);
"""
error = """
Error while reading queries [path: "queries/module_1.sql"]:  --> 2:45
  | 
  | INSERT INTO Author (id, name) VALUES (:id, $1);
  |                                             ^---
  | 
  = Indexed bind parameters (`$index`) are not allowed when using the extended syntax.
  = Consider using a named bind parameter like `:identifier`, or use the PostgreSQL-compatible syntax..
"""

[[test]]
name = "ColumnNameAlreadyTaken"
migration = """
CREATE TABLE item (
  name TEXT,
  price INTEGER
);
"""
query = """
--! items
SELECT name, price AS name FROM item;
"""
error = """
Error while preparing query "items" [file: "queries/module_1.sql", line: 1]:
Two or more columns have the same name: `name`. Consider disambiguing the column names with `AS` clauses.
"""
```
Then we have a test runner that generates a temporary file tree with the specified queries and migrations, runs cornucopia CLI on it and checks the returned error. By testing at the boundaries, we can change the underlying code without changing the tests and when we change the error message, we should have an option for the test runner to automatically update the error message in the test descriptor.

Virgiel:
The test runner is a separate binary that can be run as a CLI with formatted colored output or as a cargo unit test. Two types of tests are available, error tests that check that errors are caught and return a well-formatted message, and example tests that check that all the examples work. 

Error tests are described in TOML files:
```toml
[[test]]
name = 'ColumnNameAlreadyTaken'
query = '''
--! items
SELECT name, price AS name FROM item;
'''
migration = '''
CREATE TABLE item (
  name TEXT,
  price INTEGER
);
'''
error = '''
Error while preparing query "items" [file: "queries/module_1.sql", line: 1]:
Two or more columns have the same name: `name`. Consider disambiguing the column names with `AS` clauses.'''
```

Closes #60

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,6 +30,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: test
+          args: --all
 
   fmt:
     name: Rustfmt
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -480,7 +480,9 @@ name = "integration"
 version = "0.1.0"
 dependencies = [
  "clap",
+ "cornucopia",
  "owo-colors",
+ "postgres",
  "serde",
  "tempfile",
  "toml",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -479,6 +479,7 @@ dependencies = [
 name = "integration"
 version = "0.1.0"
 dependencies = [
+ "clap",
  "owo-colors",
  "serde",
  "tempfile",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -479,6 +479,7 @@ dependencies = [
 name = "integration"
 version = "0.1.0"
 dependencies = [
+ "owo-colors",
  "serde",
  "tempfile",
  "toml",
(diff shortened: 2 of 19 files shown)


Virgiel:
We could also move the codegen tests to the new binary. Having them in the Cornucopia crate is annoying when experimenting.

LouisGariepy:
@Virgiel Thanks for your work on this PR. Now that #52  is out of the way, I'll be able to look into it soon.

Do you think it'd be useful if I implemented #55 sooner rather than later so you can use the lib API?

Virgiel:
> Do you think it'd be useful if I implemented #55 sooner rather than later so you can use the lib API?

Yes, this would allow testing the current codegen change without reinstalling the binary on the system

Virgiel:
We should run the integration tests and examples in the CI

LouisGariepy:
Nice work. 

> We could also move the codegen tests to the new binary. Having them in the Cornucopia crate is annoying when experimenting.

What's your idea for codegen tests? Do you think it'd be sufficient to match the generated code string? Or should we actually generate some usage code and actually do a insert/select assertion, as was done previously?

LouisGariepy:
> We could also move the codegen tests to the new binary. Having them in the Cornucopia crate is annoying when experimenting.

Agreed. Do you reckon we should match the generated code string, or should we

Virgiel:
As long as the generated code compiles, everything should be fine since `postgres` handles all the actual logic. We should still have some usage code to stress the lifetime statements.
We would need to test both async and sync, and have a fairly complex setup to touch every part of the codegen code.

LouisGariepy:
A massive improvement. Sorry it took a bit of time to review. Looks good to me.

PR #64, opened 2022-05-24
Remove deadpool

Virgiel:
While connection polling is useful for server use, it is useless for code generation. I would even favor using `postgres` instead of `tokio-postgres` for simplification.


Virgiel:
If you want, you can merge it as is. 
I will remove the use of async in codegen in a future PR. I was working on it today and got stuck on the need to run migrations in both async tests and synchronous code generation. When we support `postgres` codegen, we will be able to run tests without async.

PR #66, opened 2022-05-24
postgres support

Virgiel:
- Added support for `postgres` codegen in addition to `tokio-postgres` codegen.
- Removed all async from cornucopia
- No longer generates a query structure for execution and returns the number of modified rows 
- Small cleanups and simplifications

It's getting too big for a single PR, so I'll just clean up some things and call it a day for now. The current state of the codegen makes it difficult to implement the new features I've proposed in other issues, I'll try to improve it in other PRs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -163,11 +163,12 @@ dependencies = [
  "clap",
  "cornucopia_client",
  "eui48",
- "futures",
+ "fallible-iterator",
  "heck",
  "indexmap",
  "pest",
  "pest_derive",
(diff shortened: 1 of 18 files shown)


LouisGariepy (changes requested):
Awesome PR!

Big kudos for the codegen refactor, I left it in a mess, but now its much better, and it'll keep getting better. This project would never have been as good as it is without your help :+1:

LouisGariepy (approved):

PR #67, opened 2022-05-24
55 expose a library equivalent to the cli

### LINKED ISSUE #55: Expose a library equivalent to the CLI.

LouisGariepy:
Closes #55 .

@Virgiel Do you think more works need to be done to make it compatible with your other PRs? Right now since this is built off of main, it still uses the old async stuff.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,17 +12,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-recursion"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2cda8f4bcc10624c4e85bc66b3f452cca98cfa5ca002dc83a16aad2367641bea"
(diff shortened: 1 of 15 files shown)


Virgiel:
If I could reuse the same connection for multiple codegen, that would be wonderful

LouisGariepy:
Should be good

LouisGariepy:
I ended up keeping `generate_managed` (creates a `cornucopia_conn` internally) as a convenience function in the API, while `generate_live` now accepts a `Client` so we can reuse the latter for multiple codegens. 

I also made public few other items (`startup_container` and `cleanup_container`) that I think could be useful for reusing connections.

PR #69, opened 2022-05-25
Fix codegen and examples

Virgiel:
- Make examples work again
- Export `postgres-types` from `cornucopia-client`


LouisGariepy:
Shoot I forgot the examples before merging yesterday, sorry!

PR #70, opened 2022-05-25
Better codegen

Virgiel:
In my quest to improve code generation, I saw that we were doing a lot of string allocations. While we are not very performance sensitive, I would still prefer to limit them before it gets out of hand. 
I managed to replace most of the `format!` with `write!` while simplifying some of the logic.

### DIFF
--- a/codegen_test/src/cornucopia_async.rs
+++ b/codegen_test/src/cornucopia_async.rs
@@ -71,13 +71,7 @@ pub mod types {
                 }
             }
         }
-        #[derive(Debug, Clone)]
-        pub struct CustomCompositeParams<'a> {
-            pub wow: &'a str,
-            pub such_cool: i32,
-            pub nice: super::super::types::public::SpongebobCharacter,
-        }
-        impl<'a> postgres_types::ToSql for CustomCompositeParams<'a> {
+        impl<'a> postgres_types::ToSql for CustomCompositeBorrowed<'a> {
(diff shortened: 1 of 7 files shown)


Virgiel:
Ideally, we should create our own print macro that unpacks errors, because writing to a string should never fail

Virgiel:
There are less lines in codegen.rs than before postgres support!

Virgiel:
To allow zero cost params structure now every domain type generates an owned, borrowed and params version. The borrowed version is `FromSql` and the params version is `ToSql`. This allows to have ArrayIterator when borrowing but slice in params. Currently those two versions can be redundant when there is no array in the type.

I try to rely more on the `postgres` macros to derive `FromSql` and `ToSql`. It's a shame that these macros don't natively support referencing, we have a lot of redundant code just to deal with this limitation. If we can implement it so easily, it shouldn't be difficult to implement it upstream.

LouisGariepy:
Looking at the history, it's been proposed at least twice, with this PR looking close to merging https://github.com/sfackler/rust-postgres/pull/574, but totally inactive since 2020. I don't know much about serde unfortunately so it'd be a bit hard for me to step in a revive this PR. It is a shame. I really didn't like the way I had to write the impls, but here we are :|

Thanks for your work on this PR, codegen is so much better.

Virgiel:
Ready for a merge!

Issue #65, opened 2022-05-24
Null everywhere 


Virgiel:
Currently, we can declare the nullity of the returned rows fields:

```sql
CREATE TYPE composite AS (
    name TEXT,
    age INT NOT NULL
);

CREATE TABLE items (
    composite composite,
    name TEXT NOT NULL,
    data BYTEA,
    datas BYTEA[]
);
```

```sql
--! items ?{composite, data, datas}
SELECT * FROM items;
```

This is enough to cover the nullity of `data`, `datas` and `composite` but not the values of `datas` and `composite.name`.

## Nullable array value

We need a way to declare the nullity of the array and the values of the array independently which current syntax does not allow.

We could get back to something close to the old syntax:
- `datas?` means nullable array
- `datas[?]` means nullable array values
- `datas?, datas[?]` or `datas?[?]` means nullable array and array values
```sql
--! items {composite?, data?, datas?[?]}
SELECT * FROM items;
```

# Nullable composite fields

We could declare it in functions declaration:

- `composite?` means nullable composite
- `composite.name?` means nullable composite's name
- `composite?, composite.name?` or `composite?.name?` means nullable composite and composite's name 

```sql
--! items {composite?.name?, data?, datas?}
SELECT * FROM items;
```

But this implies a redeclaration each time we use this composite type. We could add a new syntax to declare the nullity of fields of database types :

```sql
--:db composite{name?}
```

# Nullable parameters

We could use the same syntax for parameters and rows:

```sql
--! new_items (composite?, data?, datas?[?]) {data?}
INSERT INTO items (composite, name, data, datas) 
  VALUES (:composite, :name, :data, :datas) 
  RETURNING name, data;
```

# Nullable named generated types fields

Depends on #63 

We could also use this syntax for named generated type :
```sql
--:params ItemParams{composite?, data?, datas?[?]}
--:row Item{composite?, data?, datas?[?]}

--! new_items (IdParams) -> Item
INSERT INTO items (composite, name, data, datas) 
  VALUES (:composite, :name, :data, :datas) 
  RETURNING *;
```

Virgiel:
Can I start working on it?

Virgiel:
I am already working on #63. However, I am not very familiar with `pest`'s grammar, so I would appreciate your help with this!

LouisGariepy:
I'll update and provide a copy here as soon as possible.

LouisGariepy:
[...]

PR #73, opened 2022-05-27
Support named generated type

### LINKED ISSUE #63: Reuse generated types
Currently, if I have two queries returning the same thing, we generate two different row types. I would like the possibility to declare that two queries return the same row type:

```sql
--! authors Author
SELECT * FROM Author;

--! authors_by_id Author
SELECT * FROM Author WHERE id = :id;
```

Should generate a single row struct: 
```rs
pub struct AuthorBorrowed<'a> {
    pub id: i32,
    pub name: &'a str,
    pub country: &'a str,
}
#[derive(Debug, Clone, PartialEq)]
pub struct Author {
    pub id: i32,
    pub name: String,
    pub country: String,
}
```

And we generate a single query type but two query functions:
```rs
pub fn authors<'a, C: cornucopia_client::GenericClient>(
    client: &'a C,
) -> AuthorQuery<'a, C, Author, 0> {
    AuthorQuery {
        client,
        params: [],
        query: "SELECT * FROM authors",
        mapper: |it| Author::from(it),
    }
}

pub fn authors_by_id<'a, C: cornucopia_client::GenericClient>(
    client: &'a C,
    id: &'a i32
) -> AuthorQuery<'a, C, Author, 1> {
    AuthorQuery {
        client,
        params: [id],
        query: "SELECT * FROM authors WHERE id = :id",
        mapper: |it| Author::from(it),
    }
}

pub struct AuthorQuery<'a, C: cornucopia_client::GenericClient, T, const N: usize> {
    client: &'a C,
    query: &'static str,
    params: [&'a (dyn tokio_postgres::types::ToSql + Sync); N],
    mapper: fn(AuthorBorrowed) -> T,
}
impl<'a, C, T, const N: usize> AuthorQuery<'a, C, T, N>
where
    C: cornucopia_client::GenericClient,
{
    pub fn map<R>(self, mapper: fn(AuthorBorrowed) -> R) -> AuthorQuery<'a, C, R, N> {
        AuthorQuery {
            client: self.client,
            params: self.params,
            query: self.query,
            mapper,
        }
    }
    pub fn extractor(row: &tokio_postgres::row::Row) -> AuthorBorrowed {
        AuthorBorrowed {
            id: row.get(0),
            name: row.get(1),
            country: row.get(2),
        }
    }
    pub async fn stmt(&self) -> Result<tokio_postgres::Statement, tokio_postgres::Error> {
        self.client.prepare(self.query).await
    }
    pub async fn one(self) -> Result<T, tokio_postgres::Error> {
        let stmt = self.stmt().await?;
        let row = self.client.query_one(&stmt, &self.params).await?;
        Ok((self.mapper)(Self::extractor(&row)))
    }
    pub async fn vec(self) -> Result<Vec<T>, tokio_postgres::Error> {
        self.stream().await?.try_collect().await
    }
    pub async fn opt(self) -> Result<Option<T>, tokio_postgres::Error> {
        let stmt = self.stmt().await?;
        Ok(self
            .client
            .query_opt(&stmt, &self.params)
            .await?
            .map(|row| (self.mapper)(Self::extractor(&row))))
    }
    pub async fn stream(
        self,
    ) -> Result<
        impl futures::Stream<Item = Result<T, tokio_postgres::Error>>,
        tokio_postgres::Error,
    > {
        let stmt = self.stmt().await?;
        let stream = self
            .client
            .query_raw(&stmt, cornucopia_client::slice_iter(&self.params))
            .await?
            .map(move |res| res.map(|row| (self.mapper)(Self::extractor(&row))));
        Ok(stream.into_stream())
    }
}
```

```rs
pub async fn usage(client: &impl GenericClient) {
    let all: Vec<Author> = authors(client).vec().await.unwrap();
    let specific: Author = authors_by_id(client, &34).one().await.unwrap();
}
```

Same idea for parameters struct:

```sql
--! book_filter(FilterParams)
SELECT * FROM Author WHERE id = :id AND name = :name;

--! authors_filter(FilterParams)
SELECT * FROM Book WHERE name = :name AND id = :id;
```

Should generate a single parameters type:
```rs
#[derive(Debug, Clone)]
pub struct FilterParams<'a> {
    pub id: i32,
    pub name: &'a str
}

impl <'a> FilterParams<'a> {
    pub fn book_filter<'a, C: cornucopia_client::GenericClient>(
        &'a self,
        client: &'a C,
    ) -> BookByIdQuery<'a, C, AuthorById> {
        book_by_id(client, &self.id, &self.name)
    }

    pub fn authors_filter<'a, C: cornucopia_client::GenericClient>(
        &'a self,
        client: &'a C,
    ) -> AuthorByIdQuery<'a, C, AuthorById> {
        author_by_id(client, &self.name, &self.id)
    }
}
```

The first time named generated types are used, we should to store their compositions somewhere and check that they match the expected compositions when they are next used.
I do not have strong opinion on the syntax.

Virgiel:
Close #63

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1,6 +1,7 @@
 use std::fmt::Display;
 
 use error::Error;
+use heck::ToUpperCamelCase;
 use pest::{iterators::Pair, Parser};
 use pest_derive::Parser as Pest;
 
@@ -70,7 +71,7 @@ impl<T: Ord> Ord for Parsed<T> {
 }
 
 impl<T> Parsed<T> {
-    pub(crate) fn map<U>(&self, f: fn(&T) -> U) -> Parsed<U> {
+    pub(crate) fn map<U>(&self, f: impl Fn(&T) -> U) -> Parsed<U> {
         Parsed {
             value: f(&self.value),
             start: self.start,
@@ -178,25 +179,54 @@ impl FromPair for Query {
 }
 
 #[derive(Debug)]
-pub(crate) enum QueryDataStructure {
+pub(crate) enum QueryDataStruct {
     Implicit { idents: Vec<Parsed<String>> },
     Named(Parsed<String>),
 }
 
-impl Default for QueryDataStructure {
+impl QueryDataStruct {
+    pub(crate) fn name_and_fields(
+        self,
+        registered_structs: &[TypeDataStructure],
+        query_name: &Parsed<String>,
+        name_suffix: Option<&str>,
+    ) -> (Vec<Parsed<String>>, Parsed<String>) {
+        match self {
+            QueryDataStruct::Implicit { idents } => (
+                idents,
+                query_name.map(|x| {
+                    format!(
+                        "{}{}",
+                        x.to_upper_camel_case(),
+                        name_suffix.unwrap_or_default()
+                    )
+                }),
+            ),
+            QueryDataStruct::Named(name) => (
+                registered_structs
+                    .iter()
+                    .find_map(|it| (it.name == name).then(|| it.fields.clone()))
+                    .unwrap_or_default(),
+                name,
+            ),
+        }
+    }
+}
+
+impl Default for QueryDataStruct {
     fn default() -> Self {
         Self::Implicit { idents: Vec::new() }
     }
 }
 
-impl FromPair for QueryDataStructure {
+impl FromPair for QueryDataStruct {
     fn from_pair(pair: Pair<Rule>) -> Self {
         let pair = pair.into_inner().next().unwrap();
         match pair.as_rule() {
-            Rule::ident => QueryDataStructure::Named(Parsed::<String>::from_pair(pair)),
+            Rule::ident => QueryDataStruct::Named(Parsed::<String>::from_pair(pair)),
             Rule::field_list => {
                 let idents = pair.into_inner().map(Parsed::<String>::from_pair).collect();
-                QueryDataStructure::Implicit { idents }
+                QueryDataStruct::Implicit { idents }
             }
             _ => {
                 unreachable!()
@@ -208,19 +238,19 @@ impl FromPair for QueryDataStructure {
 #[derive(Debug)]
 pub(crate) struct QueryAnnotation {
     pub(crate) name: Parsed<String>,
-    pub(crate) param: QueryDataStructure,
-    pub(crate) row: QueryDataStructure,
+    pub(crate) param: QueryDataStruct,
+    pub(crate) row: QueryDataStruct,
 }
 
 impl FromPair for QueryAnnotation {
     fn from_pair(pair: Pair<Rule>) -> Self {
         let mut tokens = pair.into_inner();
         let name = Parsed::<String>::from_pair(tokens.next().unwrap());
-        let (mut param, mut row) = <(QueryDataStructure, QueryDataStructure)>::default();
+        let (mut param, mut row) = <(QueryDataStruct, QueryDataStruct)>::default();
         for it in tokens {
             match it.as_rule() {
-                Rule::query_param => param = QueryDataStructure::from_pair(it),
-                Rule::query_row => row = QueryDataStructure::from_pair(it),
+                Rule::query_param => param = QueryDataStruct::from_pair(it),
+                Rule::query_row => row = QueryDataStruct::from_pair(it),
                 _ => {
                     unreachable!()
                 }
--- a/src/prepare_queries.rs
+++ b/src/prepare_queries.rs
@@ -1,7 +1,5 @@
-use std::rc::Rc;
-
 use crate::{
-    parser::{Parsed, QueryDataStructure, TypeDataStructure},
+    parser::{Parsed, TypeDataStructure},
     read_queries::ModuleInfo,
     type_registrar::CornucopiaType,
     type_registrar::TypeRegistrar,
@@ -10,12 +8,14 @@ use crate::{
 };
 use error::Error;
 use error::ErrorVariant;
-use heck::ToUpperCamelCase;
+
 use indexmap::{map::Entry, IndexMap};
 use postgres::Client;
 use postgres_types::{Kind, Type};
+use std::rc::Rc;
 
-/// This data structure is used by Cornucopia to generate all constructs related to this particular query.
+/// This data structure is used by Cornucopia to generate
+/// all constructs related to this particular query.
 #[derive(Debug, Clone)]
 pub(crate) struct PreparedQuery {
     pub(crate) name: String,
@@ -42,7 +42,7 @@ pub(crate) struct PreparedParams {
     pub(crate) queries: Vec<usize>,
 }
 
-/// A returned row
+/// A returned row struct
 #[derive(Debug, Clone)]
 pub(crate) struct PreparedRow {
     pub(crate) name: Parsed<String>,
@@ -93,7 +93,6 @@ impl PreparedModule {
         match self.rows.entry(name.value.clone()) {
             Entry::Occupied(o) => {
                 let prev = &o.get();
-
                 // If the row doesn't contain the same fields as a previously
                 // registered row with the same name...
                 validation::named_struct_field(
@@ -299,44 +298,22 @@ fn prepare_query(
     registrar: &mut TypeRegistrar,
     param_types: &[TypeDataStructure],
     row_types: &[TypeDataStructure],
-    query: ValidatedQuery,
-) -> Result<(), Error> {
-    // Prepare the statement
-    let stmt = client
-        .prepare(&query.sql_str)
-        .map_err(|e| Error::new(e, &query.name, module.info.clone()))?;
-
-    let ValidatedQuery {
+    ValidatedQuery {
         name,
         params,
         bind_params,
         row,
         sql_str,
-    } = query;
+    }: ValidatedQuery,
+) -> Result<(), Error> {
+    // Prepare the statement
+    let stmt = client
+        .prepare(&sql_str)
+        .map_err(|e| Error::new(e, &name, module.info.clone()))?;
 
-    let (nullable_params_fields, params_name) = match params {
-        QueryDataStructure::Implicit { idents } => {
-            (idents, name.map(|x| x.to_upper_camel_case() + "Params"))
-        }
-        QueryDataStructure::Named(name) => (
-            param_types
-                .iter()
-                .find_map(|it| (it.name == name).then(|| it.fields.clone()))
-                .unwrap_or_default(),
-            name,
-        ),
-    };
-    let (nullable_row_fields, row_name) = match row {
-        QueryDataStructure::Implicit { idents } => (idents, name.map(|x| x.to_upper_camel_case())),
-        QueryDataStructure::Named(name) => (
-            row_types
-                .iter()
-                .find_map(|it| (it.name == name).then(|| it.fields.clone()))
-                .unwrap_or_default(),
-            name,
-        ),
-    };
-
+    let (nullable_params_fields, params_name) =
+        params.name_and_fields(param_types, &name, Some("Params"));
+    let (nullable_row_fields, row_name) = row.name_and_fields(row_types, &name, None);
     let params_fields = {
         let stmt_params = stmt.params();
         let params = bind_params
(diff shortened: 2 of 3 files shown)


LouisGariepy:
I'm a bit short on time this weekend, but I am working on it! The grammar changes are relatively small, most of the work will be done in the parser.

Virgiel:
No problem, we are not in a hurry 😁 I am preparing the codegen change

Virgiel:
@LouisGariepy  CI is a bit slow, have you tried enabling [caching](https://github.com/actions/cache/blob/main/examples.md#rust---cargo) ?

LouisGariepy:
> @LouisGariepy CI is a bit slow, have you tried enabling [caching](https://github.com/actions/cache/blob/main/examples.md#rust---cargo) ?

I will open an issue and look into fixing it soon :+1:

LouisGariepy:
I'm almost done, but I'm still pondering some details. The suggested syntax for named parameters is not valid for the pg-compatible syntax. Unfortunately I can't think of a syntax that works that isn't obscure or cumbersome :frowning_face: Maybe you have a better idea.

If there's no clear solution, I suggest we just go forward and restrict this PR to the extended syntax.

Virgiel:
You are right. I have some ideas on how this could work, but it will lead to errors. Let's stick to the extended syntax for now. 

I would also remove the use of indexes to declare nullable columns, although this method is less verbose, I find it more error prone and the verbosity is mitigated by named types. One of the aims of this library is precisely to not use indexes ourselves.

LouisGariepy:
@Virgiel Agreed! I will proceed with that in mind.

LouisGariepy:
I think the parsing part (including errors) is mostly ready. Tests fail because `the name `AuthorNameById` is defined multiple times`, but I believe this is related to codegen. I might be wrong though. Please let me know if I overlooked something :smiley: or if more work is needed.

LouisGariepy:
Thanks for the hotfix :)

LouisGariepy:
I think the parsing part (including errors) is mostly ready, bar some code improvements. Please let me know if I overlooked something, or if more work is needed :smiley:.

Virgiel:
I think a little bit about grammar, we currently have something like this:

```sql
--! query Params Row
--! query () Row
```

It works most of the time but can be ambiguous:
```sql
--! query Name 
```

Perhaps we could use a syntax similar to that of the Rust functions:
```sql
--! query Params
--! query -> Row
--! query Params -> Row
```

When merging with other proposals (#76 and #65), such as params and column with contains only names we could use the same grammatical rules for both. It would look something like this:
```sql
--! query (first?, second?, third)
--! query -> {first?, second?}
--! query (first?, second?, third) -> {first?, second?}
```


We could use `:` instead of `->`, it is arguably cleaner.:
```sql
--! query (first?, second?, third)
--! query : {first?, second?}
--! query (first?, second?, third) : {first?, second?}
```

We could use `()` for parameters and columns, but as parameters can contain non-nullable names when using Postgres syntax and columns can only contain nullable names, this could be confusing. However, I think it would make the grammar much simpler:
```sql
--! query (first?, second?, third)
--! query : (first?, second?)
--! query (first?, second?, third) : (first?, second?)
```

LouisGariepy:
On it!

LouisGariepy:
> [...] as parameters can contain non-nullable names when using Postgres syntax and columns can only contain nullable names, this could be confusing.

Sorry I'm not sure I understand this part. But I'm going to work the grammar and parser to accept
```sql
--! query (first?, second?, third)
--! query : (first?, second?)
--! query (first?, second?, third) : (first?, second?)
```

Virgiel:
Columns are always named implicitly and we use the declaration only to declare the nullity:
```sql
--! query : {col2?}
SELECT col1, col2 FROM table;
```
The same applies to parameters when we use the extended syntax : 
```sql
--! query(col2?)
SELECT * FROM table WHERE col1=:col1 AND col2=:col2;
```
However, with Postgres syntax, the declared parameters give the parameter names in order and their nullity:
```sql
--! query(col1, col2?)
SELECT * FROM table WHERE col1=$1 AND col2=$2;
```

This is why I think the use of `()` for parameters and columns can be confusing for users as they don't always mean the same thing.

LouisGariepy:
@Virgiel Gotcha! Thanks for explaining.

> `{}` is always order-independent and `()` is order-independent sometimes (i.e. except in parameter position for pg-compatible syntax).

Vs

>`()` is order-independent except in parameter position for pg-compatible syntax.

It seems to me like the first option adds complexity, but maybe I'm missing the mark completely. Sorry if I'm wasting your time.

Also, just to confirm, are you suggesting a grammar that accepts something like this?: `--! query_name ParamStructName(param1?, param2) : RowStructName(col1, col2?)`. If so, we might as well also implement named param/row structs for the pg-compatible syntax too.

Virgiel:
> It seems to me like the first option adds complexity, but maybe I'm missing the mark completely. Sorry if I'm wasting your time.

Actually, you are right, and simple syntax is always easier to teach. I just wanted to make sure we understood the choice we were making.

Virgiel:
For this PR, I wanted to support the implicit syntax :
```sql
--! query(col1, col2?): (col1)
```
and the named one:
```sql
--! query Params: Row
```
The aim was to allow their reuse with other queries:
```sql
--! query Params: Row
--! query2 Params: (col1)
--! query3 : Row
```
But this means that the nullity has to be declared somewhere, so I have proposed an new syntax for this in #55. This can also be used to declare columns for use with the Postgres syntax:
```sql
--:params Params(composite?, data?, datas?[?])
--:row Row(composite?, data?, datas?[?])
```

What you propose is interesting because it would allow the user to choose a custom name for parameters and rows, even without reuse. But do you want it to also allow declaring reusable types?

To summarise:

# Separate declaration

```sql
--:params Params(col, composite?, data?, datas?[?])
--:row Row{composite?, data?, datas?[?]}
--! query Params: Row
```

-additional syntax
+separation of concerns

# Inline declaration

```sql
--! query Params(col, composite?, data?, datas?[?]): Row(composite?, data?, datas?[?])
--! query2 Params: Row
```

+blend into current syntax
-mixed concern

# Hybrid declaration

We can use the separate declaration for reused named types and the inline declaration to rename the generated types for a single query.

Virgiel:
What I like about your inline declaration proposal is that it blends elegantly into the current syntax, but I can see how it can be confusing when trying to understand how and where types are declared.

LouisGariepy:
Thank you for your summary, the issue is now clearer to me. After thinking this through, I think separate declaration is best.

LouisGariepy:
Made some progress on the grammar (in line with yesterday's discussion).

Virgiel:
codegen now support extended nullability (null array values and, null params, row and composite fields)

While this works, I'm still having trouble figuring out where we should put the nullability metadata for composite types and arrays, but I have some ideas. We could generate a nullability metadata tree instead of storing them into `PreparedField`, that way we can cleanly support layers of nullability like `data?[?]` or even `composite?.col?`.


Should the nullability declaration for composite types be applied to all modules (which is what happens now), per module or per query? If we allow the per-module declaration or even the ultimate `composite.col?` syntax, we will generate multiple composite types that can't always be shared between queries or modules.

LouisGariepy:
Per-module nullability is weird because it implies that you have intimate knowledge of your data and how its nullability is affected by the queries in this particular module vs other modules (again hard to debug/refactor, but maybe I'm not seeing the full picture here).

But if the nullability is shared between modules, I could see it being hard to debug and refactor when you don't know where a particular statement is located. Longshot, but if type definition are shared, maybe we should have them live in a separate file/folder to emphasize that fact and ease conflict resolution / error reporting.

In any case, I'm leaning towards shared nullability.

LouisGariepy:
Also, just to be sure, does this PR officially include #65 and #73 in its scope?

Virgiel:
I prepared the codegen for #65 and already found some bugs in the code because of that, but we can enable it in another PR and focus on the generated type named for this PR

LouisGariepy:
> This can also be used to declare columns for use with the Postgres syntax:
>
> --:params Params(composite?, data?, datas?[?])

If we allow the query styles (pg and extended) to be mixed within a run of cornucopia, we have to require that `--: params` annotations spell out all parameters in order (even non-nullable ones, even when using the extended syntax). At least that's my understanding when you write
```sql
--:params Params(col, composite?, data?, datas?[?])
                                ^--- This wouldn't have to be there in extended syntax
```

At this point I'm not too attached to the pg-compatible syntax and I'm wary to introduce too many edge cases because of it. Here are some possible solutions

# 1. `--syntax` flag
User must chose a syntax style when using Cornucopia. This would make it easier to completely isolate the grammar and parser, errors, etc. It would also make users more aware of what syntax they should be using (less chance mix up the two styles).

# 2. Require that type annotations spell out columns in order all the time.
This would make the syntax homogenous for postgres/extended syntaxes and also for params/rows. Obviously this introduces some unnecessary verbosity in some cases.

# 3. Remove postgres syntax altogether
This is drastic, but it might be worth it. I don't know if this feature is actually useful, or if it just adds unnecessary maintenance.

Please let me know what you think.

LouisGariepy:
Still working on the parser... Will have more time soon :smile:

LouisGariepy:
Made some good progress on the new parser. It's a lot cleaner modular which should help to experiment with different grammars in the future. It can also handle the new type annotations `--: ROW`, etc. I still have to work on wiring all that with `prepare_queries`, and to add new validation and error tests.

Virgiel:
I can help you with the wiring if you want, I made some recent changes to `prepare_queries` so it will be easier for me

LouisGariepy:
Sorry for being unresponsive! Fortunately, I have some time today to work on this PR. Rewriting the grammar and parser took me a while, and there was a lot of code to add to `prepare_queries` because of type annotations, nullable params, new error checks, etc. Main crate is now compilable, and most test pass, but I still have some error failures and some (known) bugs.

When tests and examples will run (later today, I hope), I'll publish my changes, and then we can do a final validation of the design and start reviewing the implementation.

Virgiel:
This is a huge job!
Looking at the code, I think we should drop the Postgres-compatible syntax. We already do a good job of making our lives more difficult, we should at least make it easier for ourselves when we can 😅

LouisGariepy:
Haha, yeah, the dual syntax does make things very awkward now that both the internals and the syntax are becoming more complex.

For the record, my interest for the pg-compatible syntax in the first place was two-fold.

Virgiel:
It works 🥳
A little rough around the edges but very promising!

Virgiel:
The parser requires a blank line between type declarations, I don't think this is necessary. Currently we require every named type to be declared, I thought that by default any undeclared named type should have all its fields non-null.

LouisGariepy:
Dropping the Pg syntax will remove some of the nastiest bits, so there's that :smile: .

> The parser requires a blank line between type declarations, I don't think this is necessary.

I implemented this syntax for defining multiple types:
```sql
--: ROW (
--:     Row1(...),
--:     Row2(...),
--:     Row3(
--:         ...
--:     )
--: )
```
Requiring the newline makes parsing simpler. We could simply stack multiple `--: row Row1 (...)`, but I thought that way would be cleaner and more flexible for users.

> Currently we require every named type to be declared, I thought that by default any undeclared named type should have all its fields non-null.

That's a sensible default. We can go with that.

Virgiel:
> We could also simply stack multiple `--: row Row1 (...)`, but I thought that way would be cleaner and more flexible for users.

I am more in favour of this. While I understand what you were trying to get at, I think adding another meaning to `()` will confuse things.

```sql
--: row Row1(...)
--: row Row2(...)
--: row Row3(
--:      ...
--: )
```

LouisGariepy:
I'll try to get these changes (remove pg-syntax, change the grammar, unknown types should be inferred as all nullable) in as soon as possible (tonight maybe, who knows!).

Thanks again for your work and help on this PR.

LouisGariepy:
Do you reckon there are any outstanding issues with this PR still (bar code quality, docs, etc.)?

Virgiel:
As our syntax becomes more and more complex, I have found that `pest` is less and less useful and we need more and more parsing logic. I looked for another library to implement parsing in Rust and I found an interesting one: [`chumsky`](https://docs.rs/chumsky/latest/chumsky/). You can find a draft reimplementation in this [branch](https://github.com/Virgiel/cornucopia/tree/new_parser). Do you think this implementation is better?

LouisGariepy:
> As our syntax becomes more and more complex, I have found that `pest` is less and less useful and we need more and more parsing logic. I looked for another library to implement parsing in Rust and I found an interesting one: [`chumsky`](https://docs.rs/chumsky/latest/chumsky/). You can find a draft reimplementation in this [branch](https://github.com/Virgiel/cornucopia/tree/new_parser). Do you think this implementation is better?

I always saw Pest as something temporary, When I started out, I chose it out of simplicity because it has a regex-like syntax, but I agree that it might not be the right tool now. While I haven't taken a deep look at your new parser, I can see how Chumsky is powerful.

Some notes: We'd probably have to refactor our error messages to look more like Chumsky's, otherwise, it might look jarring. Also, I assume this is WIP, but just in case, the sql-string part of the syntax is necessary to eliminate false positives (e.g. `$$this is not a :bind_param$$`).

All in all, I am in favor of a new parser. If you'd like, you can open a separate draft PR with your branch, I appreciate the work you put into this.

If we're going to change the parser, we might as well do it before #65 IMO.

Virgiel:
> Also, I assume this is WIP, but just in case, the sql-string part of the syntax is necessary to eliminate false positives (e.g. `$$this is not a :bind_param$$`).
I now better understand what you were doing. I will fix this tomorrow an create a PR

Virgiel:
The parsing logic is really hard. I really appreciate the work you put into it! 
I opened a PR with my new parser, it should be al least as good as the current one and even better at handling odd spaces.

Virgiel:
If this is OK for you we can merge this PR so we can work on #82 and #80 which are dependent on this one

PR #80, opened 2022-06-03
Erase domains types wrapper

Virgiel:
Our generated domain types do not enforce the domain rules and are only used for ser/deser with the Postgres protocol. By being exposed to the client in parameters and row types, they bring noise without any benefit.

## Deserialization

Deserialization is easy, we can simply skip the domain wrapper as it is not needed.

## Serialization

We rely on the fact that we generate `FromSql` ourselves to inject the wrapper at the last minute.

Depends on #73


LouisGariepy:
Looks good to me! Is this ready to merge?

Virgiel:
I'm not sure if the domain wrapper type should be in the cornucopia client or in the generated code. I'd like to keep it private as it's only used internally, but copying it every time into the codegen seems silly.

If you are happy with the current design, you can merge !

LouisGariepy:
That is indeed a good question.

I think we can make it very clear that cornucopia_client is intended to be consumed by the cornucopia's codegen, and not something users should depend on. If there are items in `cornucopia_client` that *are* intended for users (I think `ArrayIterator` is one?), then we could move them to a separate `cornucopia_utils` crate or whatever, intended for advanced users.

What do you think?

Virgiel:
Another crate may be excessive. How about a `core` module in the cornucopia client using [`doc(hidden)`](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#hidden). This way, the code can be used by our generated code but is hidden from the user. `rust-analyzer` could even [support it](https://github.com/rust-lang/rust-analyzer/issues/9911) in the future.

PR #81, opened 2022-06-07
Refactor validation error

Virgiel:
I have refactored the validation error logic to use less code while producing the same result. What do you think?

I considered using a library like [miette](https://github.com/zkat/miette), but it seems too big for our simple use.


LouisGariepy:
:+1: Awesome! I have some (orthogonal) unreleased changes on #73 which will integrate nicely with this.

Virgiel:
No longer relevant

PR #82, opened 2022-06-09
New parser

Virgiel:
Replaces the current `pest` based parser with one based on `chumsky`. The new parser is implemented in code only and handles more corner cases (see syntax.sql). It also opens the way for well-formatted parser errors.

A closing `;` is now mandatory at the end of queries.

Depends on #73


LouisGariepy:
Doesn't seem to be any regressions, parser code is modular, extensible and code-only, string escapes seem to work perfectly. I saw that annotations must be single-line now, but that's probably for the better. Good stuff!

LouisGariepy:
Agreed! Error reporting is one of the fronts on which Cornucopia can really elevate DX with Rust and SQL, so this is pretty important to me. I think we should open an issue and start looking into prior art and different approaches. 

Since our error reports are *just* decent enough, I won't block v0.8 on it (so we can start getting feedback on our recent changes), but maybe it'll be easy enough to squeeze it in before the next release.

PR #86, opened 2022-06-10
72 support for #[derive(Serialize)]

LouisGariepy:
Adds a `serialize` flag that derives `serde`'s `Serialize` on owned generated items.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -232,8 +232,6 @@ dependencies = [
  "postgres",
  "postgres-protocol",
  "postgres-types",
- "serde",
- "serde_json",
  "tokio",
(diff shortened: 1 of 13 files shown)


Virgiel:
How about [no wrapper type](https://github.com/Virgiel/cornucopia/tree/better_json) instead?

Virgiel:
The best API would be one that is generic over `Json<T>` instead of `serde_json::Value`. This would be difficult for codegen but I can work around it if we consider it useful.

LouisGariepy:
No wrapper :smile: I like that. Given the simplicity gains. I think we should go with it!

PR #90, opened 2022-06-12
Use GAT to live our Query API dream

Virgiel:
This is an experimental PR showing how the soon-to-be-stabilised[GAT](https://github.com/rust-lang/rust/pull/96709) feature can help us generate less code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -316,6 +316,7 @@ dependencies = [
  "async-trait",
  "cornucopia_client_core",
  "deadpool-postgres",
+ "futures",
  "tokio-postgres",
 ]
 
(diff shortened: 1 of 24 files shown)


LouisGariepy:
I think GATs also stabilize async in traits? A lot of improvements!

LouisGariepy:
There seems to be some uncertainty about the stabilization. I really hope it goes through :sweat_smile: .

LouisGariepy:
:tada: GATs have been merged :) Barring some further issues, I think that means we can expect them on stable in about 3 months or so.

Virgiel:
Unfortunately, this solution has reached a dead end. The orphan rule prevents us from generating the Borrow trait for types from other crates. The truth is that we don't need GAT, it would not improve performance or compile time. If we want to make the generated code smaller or easier to read, we can always use macro_rules to de-duplicate the code generation.

PR #91, opened 2022-06-12
Null everywhere

### LINKED ISSUE #65: Null everywhere

Virgiel:
Support for internal nullity for array values and database types.

Close #65

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1199,9 +1199,9 @@ dependencies = [
 
 [[package]]
 name = "uuid"
-version = "1.1.1"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c6d5d669b51467dcf7b2f1a796ce0f955f05f01cafda6c19d6e95f730df29238"
+checksum = "dd6469f4314d5f1ffec476e05f17cc9a78bc7a27a6a857842170bdf8d6f98d2f"
 dependencies = [
  "serde",
 ]
--- a/codegen_test/migrations/1648515058_domain.sql
+++ b/codegen_test/migrations/1648515058_domain.sql
@@ -3,9 +3,17 @@ CREATE DOMAIN domain_json AS JSON;
 CREATE DOMAIN domain_nb AS INT;
 CREATE DOMAIN domain_array AS domain_json[];
 
-CREATE TABLE nightmare_domain (
+CREATE TYPE domain_composite AS (
     txt domain_txt,
     json domain_json,
     nb domain_nb,
     arr domain_array
+);
+
+CREATE TABLE nightmare_domain (
+    txt domain_txt,
+    json domain_json,
+    nb domain_nb,
+    arr domain_array,
+    composite domain_composite
 );
--- a/codegen_test/migrations/1648515058_named.sql
+++ b/codegen_test/migrations/1648515058_named.sql
@@ -1,7 +1,16 @@
 -- Write your migration SQL here
-CREATE TABLE Item (
+CREATE TABLE named (
     id serial NOT NULL,
     name TEXT NOT NULL,
     price FLOAT,
     show BOOLEAN NOT NULL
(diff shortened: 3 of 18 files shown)


Virgiel:
While working on this PR I removed named type kind from the syntax:
```sql
--: row Named(price?)
--: param NamedParams(price?)
```
become
```sql
--: Named(price?)
--: NamedParams(price?)
```

These kind annotations are not necessary as named types are already prohibited from having the same name. This can add complexity to the validation logic, but it simplify the syntax. We could possibly always add `Params` when generating the params type, which would allow the following:

```sql
--: Named(price?)
```

Virgiel:
The bad news is that I also discovered that my extended test for domain type support was not used and the test is failing :/

LouisGariepy:
> While working on this PR I removed named type kind from the syntax:

+1!

> The bad news is that I also discovered that my extended test for domain type support was not used and the test is failing :/

Oof. Sorry, should've been more careful when I reviewed the PR 😳

Virgiel:
I found a fix and made the test even more extreme. There is only one place where `accepts` returns false but the database accepts `to_sql` output, so I always return to true. I'm not sure that `accepts` has any real use.

Virgiel:
We could simplify a lot by bringing back a generic domain wrapper for params only https://github.com/Virgiel/cornucopia/tree/domain_wrapper.

Virgiel:
Fix all domain related errors and test their support more severely.

Virgiel:
Ready to merge if it suits you

LouisGariepy:
:+1: Its crazy to me how far the project has come since I made it public. Thank you for you contributions!  I especially appreciate the work you put into the test suite. Domains now seem to be escaped and injected in the right places, and we have our intended nullity syntax. Amazing!

Virgiel:
Thank you! I enjoy working on this project with you. I think we are doing something that people will appreciate.
Unless you want the Statement API to be ready for v0.8, we just need to work on errors, documentation and testing 🥳

PR #95, opened 2022-06-14
Statement API

Virgiel:
The complexity of code generation increases a lot, and we have to do some acrobatics to handle async, but it works

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -228,6 +228,7 @@ dependencies = [
  "async-trait",
  "deadpool-postgres",
  "fallible-iterator",
+ "postgres",
  "postgres-protocol",
  "postgres-types",
  "tokio",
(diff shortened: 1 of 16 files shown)


LouisGariepy:
The codegen is starting to get wild again, but its not too bad, imo.

I see some `unsafe` was used for the statement cache. This is understandable, we don't want to introduce unnecessary overhead, but I'd prefer if there was at least a comment on the safety assumptions, even if they are very evident from the context.

Looking at the examples, I think the api is consistent, not too verbose but also explicit, many ways to do things, not many footguns... I think the added internal complexity might be worth it, but I also want to be sure this is the right move since this is a pretty big change. 

If you think this introduces too much complexity, we can let it simmer for a while until we have some more ideas. Otherwise, I'm OK to go forward.

Virgiel:
Now that I've got it working, I think we can send it out. We know that the API is going to undergo more changes soon (when GAT is stabilised for example), so we'll have the opportunity to roll back or improve it. 

I think we should do an alpha version soon to get people's opinions, maybe someone has an interesting tip.

LouisGariepy:
:partying_face:

PR #98, opened 2022-06-15
Add bench

### LINKED ISSUE #97: Benchmarks.
I think this could be a useful tool to detect regressions, and evaluate the benefits of performance-related improvements. My main goal is to compare with `postgres`/`tokio_postgres`, since we aim to be a thin wrapper on top of these, and not a comparison with all the other DB crates out there.

We could take some inspiration on the diesel benchmarks https://github.com/diesel-rs/diesel/tree/master/diesel_bench.

Virgiel:
Inspired by `diesel` benchmark but modified to prepare the database once per benchmark (more accurate and faster) and only against Postgres sync for the moment. Results are subject to variation but show that `cornucopia` adds a little bit of overhead.

For insertion, `diesel` generates multi-rows insertion on the fly and we can't compete with that yet.  We could add a feature for these types of bulk insertion.

On the bright side, the `cornucopia` benchmark is immensely more succinct and easy to use!

Close #97


LouisGariepy:
To find where we're losing time to `postgres`, I imagine the next step would be to generate flamegraphs.

LouisGariepy:
Tangential: 

I think the build script is messing with the CI for some reason? Honestly I'd like to make the build script somehow inactive during development because it causes my IDE to slow down and sometimes causes some crashes.

I think the crashes occur because testing will write generated files for examples, which triggers the build script, which then tries to create a container while the test container is already in use.

Virgiel:
> Honestly I'd like to make the build script somehow inactive during development because it causes my IDE to slow down and sometimes causes some crashes.

How do you want to proceed? Just put it in a comment or do you know a better way?

LouisGariepy:
I'll be happy to merge this as soon as I'm done with #104, since this is what's blocking the CI.

Great PR, as always, these reports will be very useful :partying_face:

PR #100, opened 2022-06-15
Miette

### LINKED ISSUE #83: Better error reporting.
One of cornucopia's goals is to facilitate SQL developer experience in Rust, and a huge part of that comes down to presenting digestible error messages to users. Right now, while serviceable, the error reporting suite is very limited in its capabilities, and not particularly aesthetically pleasing. We can do better.

I'd like to use this issue as a starting point for ideas and potential solutions. If you have an opinion on this matter, or have worked with these frameworks , please chime in, we'd love to hear from you :smile:

LouisGariepy:
Closes #83.

Might be some cleanup left, and some error messages could maybe tweaked a bit, but should be done overall. The error reporting should be better in pretty much every sense.

One weird thing is that `miette` takes into account the console size when generating error messages (I think?). This can lead to surprising results when running the error message tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "adler"
+version = "1.0.2"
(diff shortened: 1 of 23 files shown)


LouisGariepy:
Throwing this (https://github.com/zkat/miette/issues/97) here because its a known issue which Cornucopia will also suffer from. Let's hope it gets fixed one day :) In the meantime, it shouldn't impact our diagnostics too much.

LouisGariepy:
So I confirmed that `miette` does indeed use `terminal_size` which makes its pretty diagnostics very unreliable in automated environments. The solution is to use the debug format instead of the pretty one for tests.

PR #105, opened 2022-06-16
Better error

Virgiel:
The purpose of this PR is to add new validation checks and errors:
- Multiple declaration of a named type
- Row declaration for execute query
- Use of a named type as Row and Params

In preparation I have:
- disabled the auto build script
- re-enabled formatted errors for integration testing 
- improved the current error format (feedback is welcome)
- simplify parser and validation

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -29,7 +29,6 @@ impl Error {
 }
 
 #[derive(Debug, ThisError, Diagnostic)]
-#[diagnostic(code(cornucopia::write_output))]
 #[error("Could not write your queries to destination file `{file_path}`: ({err})")]
 pub struct WriteOutputError {
     pub(crate) file_path: String,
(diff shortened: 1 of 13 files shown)


src/error.rs
```diff
@@ -1,4 +1,4 @@
-use miette::Diagnostic;
+use miette::{Diagnostic, GraphicalReportHandler, GraphicalTheme};
```
LouisGariepy, line 1: Ahh! Perfect. I knew something like `GraphicalTheme::unicode_nocolor()` should exist, but I didn't know where to start looking.

LouisGariepy:
Solid changes all around.

Virgiel:
I added errors for :
- Multiple declarations of a named type
- Row declaration for execute query

Other checks could be added, especially we could check that generated types never cause a name confict, but this is rather difficult to do.

PR #108, opened 2022-06-20
Single column

### LINKED ISSUE #102: Don't create row type for queries returning a single column
Currently, every query returning columns has a dedicated type. For simple queries returning a single column, this creates a lot of API noise for minimal ergonomic improvements. How about we return the row type without a wrapping struct ?

```sql
--! insert
INSERT INTO table (...) VALUES (...) RETURNING id
```

Would return an `i32` instead of `Insert {id: i32}`.

LouisGariepy:
Closes #102.

Might need some cleanup (advice welcome), but overall seems to work just fine.

### DIFF
--- a/cornucopia/src/codegen.rs
+++ b/cornucopia/src/codegen.rs
@@ -30,8 +30,10 @@ impl PreparedField {
         }
     }
 
-    pub fn brw_struct(&self, for_params: bool) -> String {
-        let it = self.ty.brw_struct(for_params, self.is_inner_nullable);
+    pub fn brw_struct(&self, for_params: bool, has_lifetime: bool) -> String {
+        let it = self
+            .ty
+            .brw_struct(for_params, self.is_inner_nullable, has_lifetime);
         if self.is_nullable {
             format!("Option<{}>", it)
         } else {
@@ -39,13 +41,16 @@ impl PreparedField {
         }
     }
 
-    pub fn owning_call(&self) -> String {
-        self.ty
-            .owning_call(&self.name, self.is_nullable, self.is_inner_nullable)
+    pub fn owning_call(&self, name: Option<&str>) -> String {
+        self.ty.owning_call(
+            name.unwrap_or(&self.name),
+            self.is_nullable,
+            self.is_inner_nullable,
+        )
     }
 
     pub fn owning_assign(&self) -> String {
-        let call = self.owning_call();
+        let call = self.owning_call(None);
         if call != self.name {
             format!("{}: {}", self.name, call)
         } else {
@@ -225,7 +230,7 @@ fn gen_params_struct(
         ..
     } = params;
     let struct_fields = join_comma(fields, |w, p| {
-        gen!(w, "pub {} : {}", p.name, p.brw_struct(true))
+        gen!(w, "pub {} : {}", p.name, p.brw_struct(true, true))
     });
     let (copy, lifetime) = if *is_copy {
         ("Clone,Copy,", "")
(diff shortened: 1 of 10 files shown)


Virgiel:
The same question for params struct, do we generate a structure for parameters when there is only one param?

cornucopia/src/codegen.rs
```diff
@@ -253,9 +258,20 @@ fn gen_params_struct(
 
         let param_values = join_comma(params, |w, p| gen!(w, "&self.{}", p.name));
         let (ret_type, pre, post) = if let Some((idx, _)) = row {
+            let prepared_row = &module.rows.get_index(*idx).unwrap().1;
+            let name = prepared_row.name.value.clone();
+            let query_row_struct = if prepared_row.fields.len() == 1 {
+                prepared_row.fields[0].own_struct()
+            } else {
+                name
```
Virgiel, line 266:
```rs
let query_row_struct = if let [field] = prepared_row.fields.as_slice() {
    field.own_struct()
} else {
```

cornucopia/src/codegen.rs
```diff
@@ -355,16 +371,26 @@ fn gen_row_structs(
             )
         };
 
+        let row_struct = if fields.len() == 1 {
+            if *is_copy {
+                fields[0].own_struct()
+            } else {
+                fields[0].brw_struct(false, false)
+            }
+        } else {
```
Virgiel, line 380:
```rs
let row_struct = if let [field] = fields.as_slice() {
    field.brw_struct(false, false)
} else {
```

cornucopia/src/codegen.rs
```diff
@@ -451,34 +477,50 @@ fn gen_query_fn(
...
+        let (row_struct_name, extractor, mapper) = if fields.len() == 1 {
+            let field = &fields[0];
+            (
+                field.own_struct(),
+                String::from("row.get(0)"),
+                field
+                    .ty
+                    .owning_call("it", field.is_nullable, field.is_inner_nullable),
+            )
+        } else {
```
Virgiel, line 496:
```rs
let (row_struct_name, extractor, mapper) = if let [field] = fields.as_slice() {
      (
          field.own_struct(),
          String::from("row.get(0)"),
          field.owning_call(Some("it")),
      )
} else {
```

Virgiel (commented):
I can't push towards your branch, so here are some code suggestions

Virgiel:
In `gen_row_struct` we should not generate a row structure for single column queries as they are no longer used.
```rs 
 if fields.len() > 1 {
     // Generate row struct
```

Virgiel:
Otherwise, I like how it simplifies our generation code and its use and how you've managed to reuse the type_registrar logic so as not to blow up the complexity of the codegen!

LouisGariepy:
Thanks for the review, will look into it!

LouisGariepy:
> Or we can unwrap the row when the row type is implicit, but keep it when it is named

That'd be my first instinct. Something like "A row with only one column will not provide a wrapper type, unless you explicitly name one."

LouisGariepy:
The implementation is sketchy, but I'm pushing it so we can work off something. I think this

> Or we can unwrap the row when the row type is implicit, but keep it when it is named

works as intended for query params and rows.

Virgiel:
I would prefer pattern matching instead of length checking, as `fields[0]` is an unwrap in disguise, but we have to wait for [`let_chains`](https://github.com/rust-lang/rust/pull/94927) to stabilise first.

LouisGariepy:
I absolutely agree with you, but on the bright side, LLVM optimizes these manually checked bounds away, as you can see using a compiler tool like https://godbolt.org/. This snippet
```rust
if params.len() == 1 {
    params[0]
}
```
produces exactly the same assembly as this one
```rust
if [param] = params {
    param
}
```
when the optimization level is set to 3 (release mode).

PR #112, opened 2022-06-22
Add codegen benchmark

### LINKED ISSUE #110: Benchmark code generation time.
Could be as simple as benchmarking our `codegen_test` code generation.

Virgiel:
Close #110 
33ms on my laptop, the results are not very stable because of IO operations for reading queries files but it should be enough to catch regression


LouisGariepy:
If we ever implement embedded queries as a feature, we could use that to benchmark everything in memory. But for now this is good.

I'll go ahead and rename the `benches` folder to `usage`, to make the structure a bit clearer.

PR #116, opened 2022-06-23
Check name conflict in generated code

Virgiel:
I thought it would be difficult, but actually, with our current architecture, it was quite easy! We now catch name conflicts in the generated code before rustc, which is much better for users.

### DIFF
--- a/cornucopia/src/codegen.rs
+++ b/cornucopia/src/codegen.rs
@@ -59,21 +59,6 @@ impl PreparedField {
     }
 }
 
-// Unused for now, but could be used eventually to error on reserved
-// keywords, or support them via raw identifiers.
-#[allow(unused)]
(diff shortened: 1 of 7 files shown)


Virgiel:
I also added a check for parameter declaration in simple queries (without biding) and cleaned up some code.

After this PR, I think we are checking everything. Can you think of another missing validation?

LouisGariepy:
Solid!

> After this PR, I think we are checking everything. Can you think of another missing validation?

I think we're not checking for reserved Rust keywords. Can't think of anything else off the top of my head, but I'll try to think some more.

LouisGariepy:
> We now catch name conflicts in the generated code before rustc, which is much better for users.

Indeed, these failures would've been very confusing for users, this is really good!

PR #126, opened 2022-06-30
Remove diesel filter in benchmark

Virgiel:


LouisGariepy:
For the record, it looks to me like the other benchmarks did not use that filter? So this PR makes the benchmarks more fair?

Virgiel:
Exactly, that was the reason why diesel seemed to be twice as fast because it only queried half the rows

PR #127, opened 2022-07-11
Allow cargo to choose older version of tokio

xoac:
Why?
- allow cargo to choose best matching version of tokio. for example solana for some reason is forcing 1.14 version of tokio
- I choose 1.13.1 as minimum version because this is first not affected version by this: https://github.com/rustsec/advisory-db/blob/2e646db5085a9ed5e1a75c1f90ea663606338a8f/crates/tokio/RUSTSEC-2021-0124.md


LouisGariepy:
@Virgiel Thoughts?

LouisGariepy:
@xoac This should be fixed by #130, can you confirm that this solution works for you?

xoac:
If there is no tokio dependency in this crate then of course it works! 

Thanks! <3

PR #129, opened 2022-07-27
Allow double-colon casts

skreborn:
Previously, double-colon casts like `::type` would trigger an error. While this was avoidable by using the function-like `cast(value AS type)` syntax instead, it's a lot simpler to use the former.

I'm unfamiliar with `chomsky`, so it might very well be that the combinators I used are not the best for the task at hand, so please take a look at that. The idea was that a cast is always a double-colon (`"::"`) followed by an identifier, and an optional pair of square brackets (`"[]"`) to denote arrays.

Let me know if this needs anything else.

### DIFF
--- a/cornucopia/src/parser.rs
+++ b/cornucopia/src/parser.rs
@@ -143,6 +143,8 @@ impl Query {
     fn sql_escaping() -> impl Parser<char, (), Error = Simple<char>> {
         // https://www.postgresql.org/docs/current/sql-syntax-lexical.html
 
+        // ::bind
+        let cast = just("::").ignored();
         // ":bind" TODO is this possible ?
         let constant = none_of("\"")
             .repeated()
@@ -169,6 +171,7 @@ impl Query {
             .ignored();
 
         c_style_string
+            .or(cast)
             .or(string)
             .or(constant)
             .or(dollar_quoted)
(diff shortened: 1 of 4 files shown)


cornucopia/src/parser.rs
```diff
@@ -143,6 +143,11 @@ impl Query {
     fn sql_escaping() -> impl Parser<char, (), Error = Simple<char>> {
         // https://www.postgresql.org/docs/current/sql-syntax-lexical.html
 
+        // ::bind
+        let cast = just("::")
+            .then(ident())
+            .then(just("[]").repeated().at_most(1))
+            .ignored();
```
Virgiel, line 150: We can do something simpler by simply ignoring `::`, the other characters being ignored anyway
```rs
let cast = just("::").ignored();
```

skreborn: Makes sense, I'll make the change.

Virgiel:
Waiting for @LouisGariepy to accept and merge this change

LouisGariepy:
Good catch! This used to work with the old parser, but some things got lost when we switched to Chomsky.

Implementation looks good to me, I'll go ahead and merge. Thanks to both of you :smile:

PR #131, opened 2022-07-27
Remove migration logic

### LINKED ISSUE #128: Use schema file instead of migrations files
I do not use migrations files in my projects and by the way, I am not sure cornucopia should handle migrations. I would like to be able to pass a schema file to the CLI instead of migrations.

```
cornucopia generate --schema schema.sql
```

Virgiel:
Remove migration logic and replace it with schema arguments

Close #128


LouisGariepy:
Small things before reviewing proper: 

Examples have not been updated to reflect these changes (i.e. they still use the `migrations/` folder and the old naming scheme), which could be confusing.

I think `read_schema` is a bit misleading since the module of that name also contains other items relate to schemas (notably executing them), and the function of that name does more than just reading the migrations.

cornucopia/src/error.rs
```diff
@@ -11,9 +11,7 @@ pub enum Error {
     ValidateQueries(#[from] crate::validation::error::Error),
     Container(#[from] crate::container::error::Error),
     PrepareQueries(#[from] crate::prepare_queries::error::Error),
-    NewMigration(#[from] NewMigrationError),
-    RunMigration(#[from] crate::run_migrations::error::Error),
-    ReadMigration(#[from] crate::read_migrations::error::Error),
+    ReadMigration(#[from] crate::read_schema::error::Error),
```
LouisGariepy, line 14: `ReadSchema` instead of `ReadMigration` (c.f. my comment on the PR thread about the `read_schema` module name)

cornucopia/src/read_schema.rs
```diff
@@ -0,0 +1,107 @@
...
+
+/// Reads schema files from arg paths, if the path is a file we read it, if the path is a directory we read all .sql files it contains.
+pub fn read_schemas(client: &mut Client, paths: Vec<String>) -> Result<(), Error> {
+    for path in paths {
+        let metadata = std::fs::metadata(&path).map_err(|err| Error::Io {
+            path: path.clone(),
+            err,
+        })?;
+        if metadata.is_file() {
+            execute_schema(client, read_schema(path)?)?
```
LouisGariepy, line 56: This will try to execute *any* file, whether it is `.sql` or not. I don't think this is intended?

LouisGariepy (changes requested):

LouisGariepy:
Otherwise, looks good to me! I haven't taken a super thorough look at the documentation changes, since they are likely to be rewritten in some ways, but that also looks good to me.

cornucopia/src/read_schema.rs
Virgiel: Yes, this provide a way to force execution on a file without having to change names

cornucopia/src/load_schema.rs
```diff
@@ -0,0 +1,91 @@
...
+/// If path is a directory we load its child SQL files.
+pub fn load_schema(client: &mut Client, paths: Vec<String>) -> Result<(), Error> {
+    for path in paths {
+        let metadata = std::fs::metadata(&path).map_err(|err| Error::Io {
+            path: path.clone(),
+            err,
+        })?;
+        // List files path to load
+        let files = if metadata.is_file() {
+            vec![path]
```
LouisGariepy, line 19: Same question as before, this will load single files of *any* extension (not just `.sql`). Is that really what we want?

Virgiel: Yes, this provide a way to force execution on a file without having to change names

LouisGariepy: Oh I see "If path is a file we load its content without checking its name". That could be surprising to users, and I don't see the benefit of not checking the filename for the sake of avoiding typos, but maybe you have a usecase for it?

LouisGariepy: Either way, I won't block the PR on this, I'm just curious smile

Virgiel: You are right, we should check the name until someone complain

LouisGariepy: > Yes, this provide a way to force execution on a file without having to change names

To me, the disparity between the single file and directory behavior for this is surprising. I just want to be sure we're leaving the least amount of surprises for end users.

Virgiel: Should we error ?

Virgiel: And for directory do we ignore non .sql files or we error on them ?

LouisGariepy (approved):

Virgiel:
For me, the least confusing behaviour would be to only accept files path and do no check on filename

LouisGariepy:
@Virgiel Agreed, I'm not too intent on checking the file extension either.

We could make it so that directory paths don't check their entry's filename (but ignore non-file entries like sub directories).

Or we disallow directory paths, with clear indications that multiple schemas can still be provided.

I don't have strong opinions either way since I mostly work with a couple schemas at the very most.

Virgiel:
Users can still use `cornucopia schema migrations/*.sql` when they have multiple migrations files

integration/src/main.rs
```diff
@@ -217,7 +217,16 @@ fn run_codegen_test(
         let local_run = if let Some(run) = &suite.run {
             // Reset DB
             reset_db(client)?;
-            cornucopia::load_schema(client, vec![format!("../{}/migrations", run.path)])?;
+            let schema_path = format!("../{}/migrations", run.path);
+            let paths = if std::fs::metadata(&schema_path).unwrap().is_dir() {
+                std::fs::read_dir(&schema_path)
+                    .unwrap()
+                    .map(|it| it.unwrap().path().to_str().unwrap().to_string())
+                    .collect()
```
LouisGariepy, line 225: Should be sorted because the order is unstable (i.e. on my machine the migrations are ran in the wrong order causing crash).

integration/src/main.rs
```diff
@@ -235,7 +244,15 @@ fn run_codegen_test(
             if local_run {
                 // Reset DB
                 reset_db(client)?;
-                cornucopia::load_schema(client, vec![schema_path.into()])?;
+                let paths = if std::fs::metadata(&schema_path).unwrap().is_dir() {
+                    std::fs::read_dir(&schema_path)
+                        .unwrap()
+                        .map(|it| it.unwrap().path().to_str().unwrap().to_string())
+                        .collect()
```
LouisGariepy, line 251: Same as above.

LouisGariepy (changes requested):

LouisGariepy:
> Users can still use `cornucopia schema migrations/*.sql` when they have multiple migrations files

Oh? How does that work? The file reading function we use (`std::fs::read_to_string`) doesn't seem to handle wildcard patterns? Maybe I'm misunderstanding what you mean.

integration/src/main.rs
Virgiel: Out of curiosity, what is your file system?  We should thank your machine for helping so much to discover order problems!

Virgiel:
> Oh? How does that work? The file reading function we use (`std::fs::read_to_string`) doesn't seem to handle wildcard patterns? Maybe I'm misunderstanding what you mean.

The [shell](https://en.wikipedia.org/wiki/Glob_(programming)) is doing the actual work, every platform should have a shell syntax or command for doing the equivalent

integration/src/main.rs
LouisGariepy: I'm running Fedora 36, whose underlying filesystem is BTRFS (I think?). It is mildly exotic, which might be why we're getting such different results :stuck_out_tongue:

LouisGariepy:
Oh gotcha!

And with that, I think we're ready to merge!

LouisGariepy (approved):

PR #133, opened 2022-07-28
Better codegen and async fix

Virgiel:
I found a better way to generate code for params struct with less boilerplate code and improved usability for async variant

### DIFF
--- a/bench/usage/cornucopia_benches/generated_async.rs
+++ b/bench/usage/cornucopia_benches/generated_async.rs
@@ -399,12 +399,16 @@ pub mod queries {
                 client.execute(stmt, &[name, hair_color]).await
             }
         }
-        impl<'a, C: GenericClient>
+        impl<'a, C: GenericClient + Send + Sync>
             cornucopia_client::async_::Params<
                 'a,
                 InsertUserParams<'a>,
                 std::pin::Pin<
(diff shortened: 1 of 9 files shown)


LouisGariepy:
Interesting!

For the trait bounds, I mostly copied from https://docs.rs/tokio-postgres/latest/src/tokio_postgres/generic_client.rs.html#14-74. I don't know if there was a reason to not make the generic client `Send + Sync` in the first place, and I'll admit the possible implications are a bit over my head, but if the compiler is happy, then I guess it is pretty harmless?

LouisGariepy:
For posterity, I'll copy over your comments from the discord:

> When I tried cornucopia in production I found a strange lifetime limitation with async execute function when using params struct. > Not a big issue since it should be fixed by async trait but we might document it
> 
> Async code produce really weird error sometimes, which hide the real problem. By adding Send and Sync restriction my "lifetime > limitation" disappeared 🤷‍♂️

LouisGariepy:
Should be ready to merge unless you want to add something :smile:

Virgiel:
> I don't know if there was a reason to not make the generic client `Send + Sync` in the first place
 
This is very useful when using an asynchronous function because, as I understand it, a future is Send and Sync when all its arguments and its local value are also end and Sync.  

> I think it'd be useful to have a minimal reproducer

I will post one soon !

> Should be ready to merge unless you want to add something 😄

Ready to merge for me

PR #134, opened 2022-07-30
Separate sync async clients

LouisGariepy:
The basic groundwork for separate sync and async clients.

The specifics might be off, I tried to go about this while making the least amount of changes, but there might be a better way to go about this.


Virgiel:
I'll put cornucopia_client_core in private to make sure I don't publish it: https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field
I would be in favour of renaming the client folders to be more succinct: `sync`, `async` and `core`. Of course, we keep long names for the crates.
Otherwise I validate these changes!

Virgiel (approved):
I'll put cornucopia_client_core in private to make sure I don't publish it: https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field
I would be in favour of renaming the client folders to be more succinct: `sync`, `async` and `core`. Of course, we keep long names for the crates.
Otherwise I validate these changes!

Virgiel:
We could even use `cornucopia_sync` and `cornucopia_async` for clients name.

LouisGariepy:
Names should be good now!

I also added `publish = false` to `client_core`.

PR #159, opened 2022-09-23
fix: Escape enum members containing reserved keywords

jacobsvante:
Just stumbled upon an issue where one of the enum members of an existing postgres enum was `box`. Cornucopia happily generated the `cornucopia.rs` file, but the result was a file with a syntax error in it.

So here comes my quick-fix to that.

This PR is not meant to be merged, just demonstrating one way to resolve the issue. I'm a bit short on time so if anyone with more intimate knowledge of the codebase wants to do it properly, then please go ahead 😉

### DIFF
--- a/cornucopia/src/prepare_queries.rs
+++ b/cornucopia/src/prepare_queries.rs
@@ -10,7 +10,7 @@ use crate::{
     read_queries::ModuleInfo,
     type_registrar::CornucopiaType,
     type_registrar::TypeRegistrar,
-    validation,
+    validation::{self, KEYWORD},
 };
 
 use self::error::Error;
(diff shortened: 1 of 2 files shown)


jacobsvante:
`#[allow(non_camel_case_types)]` needs to be added to `pub mod types`, to avoid warnings.

jacobsvante:
Hmmm. It seems that postgres/tokio-postgres does not like enum with raw identifiers. Gives me the following error when I try to execute a SELECT with the given column included:

```
thread 'main' panicked at 'error retrieving column 9: error deserializing column 9: cannot convert between
the Rust type `data_import::cornucopia::types::public::Uom` and the Postgres type `uom`',
/Users/jacob/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-postgres-0.7.7/src/row.rs:151:25
```

The enum looks like this:
```rust
        pub enum Uom {
            pair,
            pcs,
            #box,
        }
```

Virgiel:
Postgres enum are [case sensitive](https://www.postgresql.org/docs/current/datatype-enum.html#id-1.5.7.15.8) which explain the error you saw.

We have two ways of dealing with these problematic cases. Our current method is to error when finding a rust keyword. We do this for tables, and we should do it for enums. It would be better to handle them elegantly by escaping them, which takes more work but would be better for usability. 

For the next version, we should cause an error, and we could start working on supporting them for a later version.

Virgiel:
Postgres enum are [case sensitive](https://www.postgresql.org/docs/current/datatype-enum.html#id-1.5.7.15.8) which explain the error you saw.

We have two ways of dealing with these problematic cases. Our current method is to error when finding a rust keyword. We do this for tables, and we should do it for enums. It would be better to handle them elegantly by escaping them, which takes more work but would be better for usability. 

For the next version, we should cause an error, and we could start working on supporting them for a later version.

jacobsvante:
Replaced by more thorough version in #161

PR #160, opened 2022-09-25
Better container error

Virgiel:
I was working on the train today and docker failed due to an unstable network. I found that our program was hiding the origin of the error and our help message was useless in this case. I simplified the code, and now we print the error message of the command.


jacobsvante:
I've hade problems with hidden error messages as well. So yay!

PR #161, opened 2022-09-25
Escape rust reserved keywords

Virgiel:
Add support for rust reserved keywords everywhere we can escape them. To achieve this we have to use an even more custom implementation of `ToSql` and `FromSql` traits, and we now have to generate them for enums.

Superseeds #159


jacobsvante:
Great job! Thank you.

Virgiel:
Ready for merge!

PR #162, opened 2022-09-29
feat: Support NUMERIC to rust_decimal::Decimal

jacobsvante:
Sorry for not adding any unit tests, stressful days. Think of this as a reminder for the time being. It seems to be working though.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,6 +42,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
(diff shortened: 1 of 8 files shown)


LouisGariepy:
Yeah we'll need some testing to ensure that everything is OK, but this is indeed the right way to support new types.

One minor concern I have is that the `ToSql` and `FromSql` impls for `rust_decimal_Decimal` are *not* maintained by the rust-postgres org. I don't think this is overly problematic though.

LouisGariepy:
Oh by the way, I can add the tests myself, no worries :smile: You can look at the diff to see how I did it if you're interested. I'll try to get to it soon.

Thanks for the PR Jacob!

jacobsvante:
Great, thanks!

PR #167, opened 2022-11-01
 Explore using macro to simplify generated code

Virgiel:
`macro_rules` should be negligible in compile time while making the generated code smaller and easier to read. This is only a demonstration as an alternative for using GAT in codegen.


jacobsvante:
Would we be able to jump to the code definition from "user code" if changing to a macro? I've found that useful many times, e.g. when I want to understand why it won't accept my parameters.

Virgiel:
It would no longer work at least until rust-analyser finds a way to jump into macro expansion. I think it is better to generate a lot of code that is fast to compile and hide nothing like we do actually

jacobsvante:
If that's the case, then I agree fully.

LouisGariepy:
>  I think it is better to generate a lot of code that is fast to compile and hide nothing like we do now

Fully agreed, especially the "hide nothing" part. It was always a goal for me to make cornucopia "just simple Rust functions", and even though its a lot more complex now, I think it still fits that mentality.

What I am concerned about is that our codegen and type_registrar modules will become so convoluted that it will become hard to maintain and to add features at a reasonable pace. For now though, I think this is solvable with more rigorous abstractions, which would keep the code simple and explicit.

In other words, I don't think that the *generated* code is too problematic, but I would like to simplify the *generating* code if possible. That'll be one of my focus areas for the next releases.

PR #169, opened 2022-11-02
Simplify codegen using template macro

Virgiel:
Compared to a simple `write!` macro the `code!` macro has many advantages:
- Use a token stream instead of a string as input, we gain improved IDE support for free
- Native support for interpolating closure which gives many opportunities to reduce allocation by lazily writing directly into the output
- Native support for repetition and reuse iterators by cloning them which gives opportunities to reuse iterators and remove our hacky `join` trick


code/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+
+/// Match allowed character at the beginning of a Rust identifier
+fn ident_start(c: char) -> bool {
+    c.is_alphabetic() || c == '_'
+}
+
+/// Match allowed character after the beginning of a Rust identifier
+fn ident_body(c: char) -> bool {
+    c.is_alphanumeric() || c == '_'
+}
```
LouisGariepy, line 19: For full compliance, this should probably use `is_xid_start` and `is_xid_continue` from https://github.com/unicode-rs/unicode-xid. AFAIK, this is how rust identifiers are defined

```rust
fn is_ident_start(c: char) -> bool {
    unicode_xid::UnicodeXID::is_xid_start(c) || c == '_'
}

fn is_ident_continue(c: char) -> bool {
    unicode_xid::UnicodeXID::is_xid_continue(c)
}
```

Note that `is_xid_start` does not include underscores (but they are valid rust identifier start), while `is_xid_continue` does include them.

cornucopia/src/codegen.rs
```diff
@@ -278,36 +245,34 @@ fn composite_fromsql(
...
-            }}
-
-            fn accepts(ty: &postgres_types::Type) -> bool {{
-                ty.name() == "{name}" && ty.schema() == "{schema}"
-            }}
-        }}"#
+                if num_fields as usize != fields.len() {
+                    return std::result::Result::Err(
+                        std::convert::Into::into(format!("invalid field count: {} vs {}", num_fields, fields.len())));
+                }
```
LouisGariepy, line 264: Is this really needed? If the number of fields is wrong, its that our codegen implementation is bugged, and thus a runtime panic is the sensible thing to do (i.e. no need to pay a runtime cost for this).

Maybe I'm wrong though, please let me know :)

LouisGariepy (commented):
A quote-like macro is a great idea :smile: Love it.

cornucopia/src/codegen.rs
Virgiel: I just follow the rust postgres macro https://github.com/sfackler/rust-postgres/blob/1e944c4ffc9e7674842fb86a8340187c731efdae/postgres-derive/src/fromsql.rs#L179.
We could argue that the generated code will always match the database schema, but I don't think these tests have a real impact on performance and may still be useful when the user changes the schema.

Virgiel: We could benchmark an improved codegen without runtime test but I propose to do this in another PR

LouisGariepy: > may still be useful when the user changes the schema.

That's a good point. hitting this may mean that our codegen is bugged (unlikely since we test our codegen) OR it might mean that the user did not regenerate their code after changing their schema.

Perhaps we could add a more explicit error message then (something along the lines of "did you forget to regenerate your query files?"). Or perhaps we could leave it as is to conform with rust postgres.

code/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+
+/// Find all non iterator interpolated identifier in `scan`
+fn ident_in_iterator<'a>(scan: &'a mut Scanner) -> Vec<&'a str> {
+    let start = scan.cursor();
+    let mut idents = Vec::new();
+    while let (_, Some(pat)) = parse_next(scan) {
+        match pat {
+            Kind::Display(ident) | Kind::Call(ident) => idents.push(ident),
+            Kind::Iterator(_) => unreachable!(),
```
LouisGariepy, line 87: This is definitely "reachable". Something along the lines of `panic!("nested repetitions are not supported")` might be clearer.

LouisGariepy:
Aside from the two small comments I left and the clippy lints, this looks good to merge :+1:

cornucopia/src/codegen.rs
Virgiel: If we change the error message here, we should change others too. I propose to leave it as is to conform with rust postgres for now and explore alternatives in another PR.

LouisGariepy:
Thanks for your work :+1: Merging!

PR #178, opened 2022-11-11
Update README.md

ricardodarocha:
Just a small contributtion
I was looking for the cli syntax and the database string connection, then I realized that this information could be here in Readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,9 +88,8 @@ You can learn more about using Cornucopia by reading our [book](https://cornucop
 The [book](https://cornucopia-rs.netlify.app/book/index.html) is the place to go to get more in-depth explanations, but here is the simplest of tasters to give you an idea.
 
 Let's say you have the following PostgreSQL queries
+**/queries/some_query_file.sql**
 ```sql
--- queries/some_query_file.sql
-
 --! authors
 SELECT first_name, last_name, country FROM Authors;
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -100,6 +99,12 @@ VALUES (:first_name, :last_name, :country)
 ```
 Notice the query annotations (`--! authors`, `--! insert_authors`) and the named bind parameters (`:first_name`, etc.).
 
+1. First of all, you can organize that queries into folder /queries/ as mentioned or pass the folder when running the CLI
```
LouisGariepy, line 102: IMO, a numbered list is not really appropriate here. Perhaps we can simply mention: "By default, Cornucopia reads all query files in the `queries/` folder".

README.md
```diff
@@ -100,6 +99,12 @@ VALUES (:first_name, :last_name, :country)
...
+2. You can generate the Rust Code with CLI, for example using a local database 
+```shell
+$ cornucopia live postgres://{usr}:{psw}@localhost:5432/{database}"
+```
```
LouisGariepy, line 106: If we're giving an example for a live database, we might as well give one for a managed one too:

"... or an ephemeral database managed by Cornucopia:
```shell
cornucopia schema {path_to_schema}
```
"

or something similar

README.md
```diff
@@ -88,9 +88,8 @@ You can learn more about using Cornucopia by reading our [book](https://cornucop
 The [book](https://cornucopia-rs.netlify.app/book/index.html) is the place to go to get more in-depth explanations, but here is the simplest of tasters to give you an idea.
 
 Let's say you have the following PostgreSQL queries
+**/queries/some_query_file.sql**
```
LouisGariepy, line 91: I think we should add "located in", as in "Let's say you have the following PostgreSQL queries located in ...".

Also, the path should be monospaced `like this` and not bolded **like that**, in my opinion (might be wrong though, I don't know much about typography).

Finally the leading slash (`/`) in the path suggests that this is an absolute path, while it really should be relative (to the cargo crate root).

README.md
LouisGariepy: The reason I did not add CLI commands to the README is that I think its better for users to read the `--help` message or the book instead, but I can see why it can be useful to get started.

LouisGariepy:
Thanks for your contribution! I hope we can make the README clearer by the end of this PR :smile:

LouisGariepy:
I have plans to significantly rework the README, both to add some more current information, and also with respect to some big incoming changes.

For this reason, I'm going to go ahead and close this PR for the time being.

PR #179, opened 2022-11-15
Use Path/PathBuf for fs paths

Virgiel:
Fix https://github.com/cornucopia-rs/cornucopia/pull/177#issuecomment-1313208321

### DIFF
--- a/cornucopia/src/lib.rs
+++ b/cornucopia/src/lib.rs
@@ -14,7 +14,7 @@ pub mod conn;
 /// High-level interfaces to work with Cornucopia's container manager.
 pub mod container;
 
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
 use postgres::Client;
 
@@ -72,7 +72,7 @@ pub fn generate_live<P: AsRef<Path>>(
 /// `podman` parameter to `true`.
 pub fn generate_managed<P: AsRef<Path>>(
     queries_path: P,
-    schema_files: &[PathBuf],
+    schema_files: &[P],
     destination: Option<P>,
     podman: bool,
     settings: CodegenSettings,
--- a/cornucopia/src/lib.rs
+++ b/cornucopia/src/lib.rs
@@ -14,6 +14,8 @@ pub mod conn;
 /// High-level interfaces to work with Cornucopia's container manager.
 pub mod container;
 
+use std::path::{Path, PathBuf};
+
 use postgres::Client;
 
 use codegen::generate as generate_internal;
@@ -41,8 +43,8 @@ pub struct CodegenSettings {
 /// set using the `settings` parameter.
 pub fn generate_live(
     client: &mut Client,
-    queries_path: &str,
-    destination: Option<&str>,
+    queries_path: &Path,
+    destination: Option<&Path>,
     settings: CodegenSettings,
 ) -> Result<String, Error> {
     // Read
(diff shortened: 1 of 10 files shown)


LouisGariepy:
I think we should use `P: AsRef<Path>` instead of `&Path` in our public APIs. For example, this makes it possible for users to supply `&str` instead of constructing a Path themselves.

I created a commit to this effect [here](https://github.com/LouisGariepy/cornucopia/commit/2f5c841cf35eafbb19b9d14948daf9957e1c6261#diff-32a5749caf9bc09241fa6e6e9b5e8fff12fb8dffa94cc3ba571fb90f6567e63a).

Otherwise LGTM.

LouisGariepy:
> This means that I can't use a String for queries_path and an `Option<PathBuf>` for destination at the same time

If we use `&Path` everywhere, we also settle on one single type for all the fields, so there's no difference in terms of expressiveness. In any case, If one path is given as some type, I assume the user will provide all paths as this type, at least for the purpose of using our API.

My goal is to keep allowing users to provide paths as `&str`, which I think is a valid usecase and very common too. It also makes this change non-breaking.

I'll leave it up to you, its not a big deal either way, but I don't really see the advantage of using `&Path` everywhere tbh.

cornucopia/src/lib.rs
```diff
@@ -70,15 +70,15 @@ pub fn generate_live(
 ///
 /// By default, the container manager is Docker, but Podman can be used by setting the
 /// `podman` parameter to `true`.
-pub fn generate_managed(
-    queries_path: &Path,
+pub fn generate_managed<P: AsRef<Path>>(
+    queries_path: P,
     schema_files: &[PathBuf],
```
LouisGariepy, line 75: `load_schema` is now defined as 
```rust
load_schema<P: AsRef<Path>>(client: &mut Client, paths: &[P]) -> Result<(), Error>
```
So we should probably also take the schema_paths as `schema_files: &[P]` instead of `schema_files: &[PathBuf]`.

examples/auto_build/_build.rs
```diff
@@ -17,7 +17,7 @@ fn main() -> Result<(), Error> {
     println!("cargo:rerun-if-changed={schema_file}");
     cornucopia::generate_managed(
         queries_path,
-        vec![schema_file.into()],
+        &[schema_file.into()],
```
LouisGariepy, line 20: If we go with the aforementioned change, we could remove that `into()` call.

LouisGariepy (commented):
Otherwise looks good.

Not related to the goal of this PR, but I appreciate the small changes you've made to avoid passing around large structs by value, or cloning potentially big strings etc. Always appreciated :)

LouisGariepy:
@Virgiel I'm curious why you prefer a separate path generic for schema paths?
```rust
<P: AsRef<Path>, F: AsRef<Path>>
``` 
I doubt this is necessary in practice, but perhaps I'm missing something? The PR is ready to be merged in any case, I'm just wondering.

Virgiel:
@LouisGariepy 
When we call `generate_managed` we have owned types `String` or `PathBuff` and owned collection `Vec<Owned>`. When we pass by reference, they become borrowed types `&str` or `&Path` and borrowed slices of owned types `&[Owned]`. So we want to be able to pass `&str` and `&[String]` at the same time and need two different generic types.

LouisGariepy:
I think you're referring to our CLI, for example, where we have PathBuf and Vec<PathBuf>. But that's fine, we can still do
```rust
generate_managed(
        queries_path, // PathBuf
        &schema_files, // Slice of PathBuf
        Some(destination), // PathBuf
        <...>
)
```
So P = PathBuf and we don't need another generic.
Passing PathBuf by value is not inefficient since its already reference-sized under the hood.

Virgiel:
> I think you're referring to our CLI, for example, where we have `PathBuf` and `Vec<PathBuf>`. But that's fine, we can still do
> 
> ```rust
> generate_managed(
>         queries_path, // PathBuf
>         &schema_files, // Slice of PathBuf
>         Some(destination), // PathBuf
>         <...>
> )
> ```
> 
> So P = PathBuf and we don't need another generic. Passing PathBuf by value is not inefficient since its already reference-sized under the hood.

You are right we can simply pass `PathBuf` by value 😅

LouisGariepy:
:100: Ready to merge

Issue #176, opened 2022-11-08
Work with both sync and async


Virgiel:
Currently, we generate either async or sync code. This is problematic if someone wants to use both in their project because types are declared twice and
therefore are incompatible.

I propose a new codegen module architecture to support both in a single file:

``` rs
pub mod types {
    // Common database declared types (Enum & composite)
    pub mod public { /* ... */ }
}
pub mod queries {
    // For each query file
    pub mod copy {
        // Common queries types (Row, Params & Borrowed)
        pub mod types { /* ... */}
        // Generated type for synchronous usage (Stmt & Query)
        pub mod sync { /* ... */ }
        // Generated type for asynchronous usage (Stmt & Query)
        pub mod tokio { /* ... */ }
    }
}
```

dbofmmbt:
I think you can work around it by generating code twice for two different destinations, e.g. 
- `src/cornucopia_sync.rs`
- `src/cornucopia_async.rs`

Actually, I think this may be a good solution for this use case. What do you think?

Virgiel:
The problem with this solution is that the database and row types are generated in both `src/cornucopia_sync.rs` and
`src/cornucopia_async.rs`. Even though they have the same name and fields, from the point of view of the Rust type system they come from different modules and are therefore incompatible, which I would like to fix

LouisGariepy:
@Virgiel I could see it being useful for perfomance reason to *choose either* sync or async for specific queries (using an *async* keyword in our annotations, for example), but I'm not sure I understand what kind of project would need to have both at the same time for all queries.

I'm not opposed, but I'd like to understand the motivation behind it.

jacobsvante:
Could be a CLI in an app that doesn't require async (and would be more cleanly written without). Then the main part of the app requires async (e.g. web server).

LouisGariepy:
@jacobsvante That would be more akin to what I suggested, wouldn't it (have some queries be sync and some be async)?

I still don't see why you would need both sync and async versions of *all* queries for that usecase.

jacobsvante:
Thanks for your thorough reply, @LouisGariepy.

I personally don't have a use case for sharing queries/types etc currently, but I can see how the synchronous CLI might want to use the same query as the async web server app (e.g. fetch all products matching the given criteria).

Virgiel:
I have a web backend where I have asynchronous endpoints and synchronous workers. I prefer to use synchronous code where I can as it is way easier to use correctly when thing get complicated. 

[...]

PR #180, opened 2022-11-15
Work with both sync and async

Virgiel:
fix https://github.com/cornucopia-rs/cornucopia/issues/176

Support generating async and sync code reusing common types.

If only async or sync code is generated we reuse the current module tree :

```rs
pub mod types {
    // Common database declared types (Enum & composite)
}
pub mod queries {
    pub mod module {
        // Common queries types (Row, Params & Borrowed)
        // Generated type for synchronous or asynchronous usage (Stmt & Query)
    }
}
```

If both sync and async code are generated we add new modules:

```rs
pub mod types {
    // Common database declared types (Enum & composite)
}
pub mod queries {
    pub mod module {
        // Common queries types (Row, Params & Borrowed)
        // Generated type for synchronous usage (Stmt & Query)
        pub mod sync { /* ... */ }
        // Generated type for asynchronous usage (Stmt & Query)
        pub mod async_ { /* ... */ }
    }
}
```

The CLI still generates async code by default and can generate async and sync when both flags are provided `--sync --async`

Even if some inner generated code has changed, this PR should not be a breaking change.


LouisGariepy:
> Should we use `r#async` or `tokio` for the async module?

As a user-facing module, my preference would be `async_` as it is a fairly standard way to disambiguate keywords. If we go for pure correctness, then `r#async` would also work.

Another way to disambiguate that is used in `rustc` is to change a letter, like `crate` -> `krate`, so `async` -> `asynk` could work, but I personally don't really like that (although I could be convinced).

To me `tokio` is misleading, it looks like it could be a module that reexports items from `tokio`.

> How should we handle sync/async combination for the CLI?

`--sync` and `--async` flags are fine for me. To generate both, you'd do both: `--sync --async`.

According to crates.io, our async client has twice as many downloads, so I think the default (when none of these flags are given) should still be only async.

LouisGariepy:
I'll try to make some time and review this (finally! :upside_down_face:). Are you still OK with this PR @Virgiel ?

Virgiel:
@LouisGariepy Ready to merge !

LouisGariepy:
Looks good to me!

PR #185, opened 2022-11-29
Allow quoted identifiers in certain places

### LINKED ISSUE #184: Dots in type names cause parse error
Dots in type names, while uncommon and generally advised against, are perfectly valid. PostgreSQL allows using quotation marks to escape names containing otherwise illegal characters.

As this is a clearly uncommon edge case, I didn't actually expect this to work, but I believe it should not be a difficult issue to fix either.

### What works

Type definitions like the following work flawlessly as-is.

```sql
--: bar(a?, b?)

-- Should be translated to "bar"
```

### What does not

But adding a dot in the name...

```sql
--: foo.bar(a?, b?)

-- Should be translated to "foo.bar"
```

...results in a parse error.

```plain
--- stderr
Error: Couldn't parse queries
```

skreborn:
Quoted identifiers are now allowed in places where it makes sense.

```sql
-- In type annotations (both name and fields).
--: "foo.bar"(a?, "b.c"?)
-- [...]

-- In query data structs (both input and output).
--! foo_bar Foo("a"?): Bar("b.c"?)
-- [...]

-- In composite type names/fields.
CREATE TYPE "named_composite.with_dot" AS (
    "this.is.inconceivable" "enum.with_dot"
);

-- In enum type names/variants.
CREATE TYPE "enum.with_dot" AS Enum('variant.with_dot');
```

When identifiers are converted to Rust names, they are normalized by replacing all non-alphanumeric characters with an underscore (`_`) in addition to escaping them with a raw identifier prefix `r#` if they happen to be a reserved keyword in Rust.

Fixes #184.

### Implementation notes

For whatever reason, I couldn't get the quoted identifier parser working with `delimited_by`, but perhaps that's just my unfamiliarity with `chumsky`. I got it to compile, but it would keep telling me it's missing a closing quotation mark, even when it was present. I managed to work around it with a number of alternative combinators, but I'll happily replace those with the proper function if anyone knows how to get it to work.


Virgiel:
Just want to add Postgres doc link as context: https://www.postgresql.org/docs/15/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

cornucopia/src/parser.rs
```diff
@@ -66,6 +66,21 @@ fn ident() -> impl Parser<char, Span<String>, Error = Simple<char>> {
...
+        .collect()
+        .map_with_span(|value: String, span: Range<usize>| Span {
+            value,
+            span: span.into(),
+        })
+}
+
+fn ident() -> impl Parser<char, Span<String>, Error = Simple<char>> {
+    plain_ident().or(quoted_ident())
+}
```
Virgiel, line 82: I think we should have this instead:
```rs
fn plain_ident() -> impl Parser<char, Span<String>, Error = Simple<char>> {
    filter(|c: &char| c.is_ascii_alphanumeric() || *c == '_')
        .repeated()
        .at_least(1)
        .collect()
        .map_with_span(|value: String, span: Range<usize>| Span {
            value,
            span: span.into(),
        })
}

fn quoted_ident() -> impl Parser<char, Span<String>, Error = Simple<char>> {
    none_of('"')
        .repeated()
        .at_least(1)
        .delimited_by(just('"'), just('"'))
        .collect()
        .map_with_span(|value: String, span: Range<usize>| Span {
            value,
            span: span.into(),
        })
}

fn ident() -> impl Parser<char, Span<String>, Error = Simple<char>> {
    plain_ident().or(quoted_ident())
}
```

Virgiel:
Sorry for pushing on your fork by mistake, I didn't know how to propose a code change on a PR.

This was the case before your PR, but we use `name` in too many places. I tried to streamline it by using an `Ident` structure: 

```rs
/// A normalized ident replacing all non-alphanumeric characters with an underscore (`_`)
/// and escaping it with a raw identifier prefix (`r#`) if it clashes with a keyword reserved in Rust.
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct Ident {
    /// Database original ident
    pub(crate) db: String,
    /// Normalized ident for rust code usage
    pub(crate) rs: String,
}
```

LouisGariepy:
Great work!

I'm unsure why we should allow it in type annotation names though. My understanding is that quoting is beneficial to refer to items that appear *in* the SQL, but type annotation names are not part of, or reffered to, by the SQL.

skreborn:
> I'm unsure why we should allow it in type annotation names though. My understanding is that quoting is beneficial to refer to items that appear _in_ the SQL, but type annotation names are not part of, or reffered to, by the SQL.

Because type names might come from the database, as well, in the form of composite types - that's what I use this syntax for most, actually.

```sql
-- Declared in schema
CREATE TYPE "foo.bar" AS ("a" float4, "b.c" float4);

-- Used in query file
--: "foo.bar"(a?, "b.c"?)
```

LouisGariepy:
Right, what I mean is that you could also write
```
--: FooBar(a?, b?)

--! awesome_query : FooBar
[...]
```
and it would work equally well. The point is that there's no *need* for the type annotation to be called `"foo.bar"` specifically, literally any name would do.

On the other hand, Cornucopia *must* be able to refer to column names exactly *as they appear in the SQL*, so being able to refer to quoted column names is a necessary addition.

I hope this explains why some identifiers *require* quoting (like column names), and some do not (like query name and type annotations).

---

Now that this distinction between required and non-required quoting is clearer, I can clarify my original question.

What's the reasoning to support quoting in type annotations when its not required? If we do allow quoting for type annotations, shouldn't we also allow it in other non-required places? Like query names, inline type names, etc?
 
For example, if we allow `--: "foo.bar"(...)` as a type annotation, shouldn't we also allow `--! "select_foo.bar"` as a query name? Otherwise it seems like an arbitrary choice to me.
 
Personally, I feel like quoting should only be allowed in required places for the sake of minimalism. But if we do allow it in non-required places, we should make sure its at least consistent (i.e. use it everywhere).

Please let me know what you think :smile:

skreborn:
If it's a return position type, yes, we could definitely get away with defining `FooBar`. But in cases where I can't define the name of the type, because it comes from the schema, I have to label them using quoted identifiers. Say there exists the following schema:

```sql
CREATE TYPE "foo.bar" AS ("a" float4, "b.c" float4);

CREATE TABLE foo (
  bar int4,
  baz "foo.bar"[]
);
```

And the following query:

```
--! query_foo
SELECT bar, baz FROM foo;
```

In this instance, I can't define the type as `FooBar`, because I have no control over the type's name - unless of course I'm missing something.

Generally, I do agree with you that we should only make it an option to quote identifiers where it's necessary, but I do think it is in this particular instance.

Let me know if I've got this all wrong, because I'm starting to feel like I'm crazy here and you know something I don't. 😅

LouisGariepy:
@skreborn Thanks for following up :)

Just so we know we're talking about the same thing, could you add the type annotation (using the new proposed quoted syntax) and its use with `query_foo`, using the same schema and everything?

skreborn:
@LouisGariepy Here goes, I hope this is what you meant, but let me know in case it's not!

First off, let's define the type so that cornucopia knows its fields are nullable. This is where the syntax addition is necessary, because I have no other way to achieve the same.

```sql
--: "foo.bar"(a?, "b.c"?)
```

This type is never mentioned again, because cornucopia knows of it and will use it whenever it encounters the `foo.bar` type coming from the database.

Afterwards, define the row type for the query. This is where no syntax changes are necessary, but technically allowed because of the previous case. Is this perhaps what you have an issue with - that this shouldn't be allowed?

```sql
--: FooQuery(bar, baz)
```

And lastly, define the query using the aforementioned row type.

```sql
--! query_foo: FooQuery
SELECT bar, baz FROM foo;
```

LouisGariepy:
@skreborn Thanks for the details, sorry for delayed explanation. I see what you mean now. Honestly this is not a usecase I had even imagined when we introduced support for named types, but I can see how this can be useful.

skreborn:
Well, I'm using it quite extensively, so I'm grateful for the feature - even if this is not what it was designed for.

Is there anything else I have to do to move this along?

LouisGariepy:
@skreborn Merging as we speak :) Thanks for your contribution.

PR #198, opened 2023-02-05
Better use of prepare and immutable bind

Virgiel:
I found a way to have explicit prepare, implicit prepare for client with cache and immutable bind:
```rs
// If can cache, cache-prepare else doesn't prepare
// Same as current API but whiteout implicit prepare when it can add unamortized latency
query().bind(&client, 1).await?;

// Prepare once and reuse statement, best perf when reused
query().prepare(&client).await?.bind(&client,1).await?; 
```
We might want to add a method to disable preparation even if it can be cached (to reduce memory cost of rare queries)

Context: https://discord.com/channels/987088069280825401/1060289763443232879


LouisGariepy:
I only did an overview, but this looks very promising from what I have seen. I'll take a full on look by the end of the week.

Thank you for you work on this PR, as always =]

LouisGariepy:
Implemented in https://github.com/halcyonnouveau/clorinde/pull/143, which has been pulled to Cornucopia in the Clorinde merge. Thank you for your contribution.

PR #205, opened 2023-03-23
Serde deserialize support

### LINKED ISSUE #203: Flag for generating Serde deserialize trait
Hi @LouisGariepy,

I have been trying out Cornucopia and am really enjoying the developer experience. Currently, I am using it as a REST API (with JSON) on top of Postgres.

I see that #86 added support for serialization of JSON, but not deserialization. I went into the generated Rust file out of curiosity and noticed that the `serde::Deserialize` trait is indeed missing. Any plans to support this? What I was thinking is that we could add a `---deserialize` switch which adds the trait, similar to the `--serialize` switch. Or, the same switch could add both traits if you feel that would be a better workflow.

My thought process is that abstractions can be created on top of SQL tables using views and this is a use-case where both serialization and deserialization would be helpful. Exposing views to an API is what [PostgREST](https://postgrest.org/en/v10.0/schema_structure.html#schema-isolation) recommends and I like the approach.

Thank you for creating this awesome project and I am looking forward to hearing your thoughts on this.

Sincerely,
Faraz

farazfazli:
Closes #203. WIP.

### DIFF
--- a/crates/cornucopia/src/codegen.rs
+++ b/crates/cornucopia/src/codegen.rs
@@ -323,7 +323,7 @@ fn gen_params_struct(w: &mut impl Write, params: &PreparedItem, ctx: &GenCtx) {
             .map(|p| p.param_ergo_ty(traits, ctx))
             .collect::<Vec<_>>();
         let fields_name = fields.iter().map(|p| &p.ident.rs);
-        let traits_idx = (1..=traits.len()).into_iter().map(idx_char);
+        let traits_idx = (1..=traits.len()).map(idx_char);
         code!(w =>
             #[derive($copy Debug)]
             pub struct $name<$lifetime $($traits_idx: $traits,)> {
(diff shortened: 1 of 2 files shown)


farazfazli:
I believe the struct fields that aren't being deserialized will have to be wrapped with Option<T>, to ensure that deserialization is performed properly. We can't use `#[serde(skip_deserializing)]` because certain types which lack a default implementation are defined in a different package.

One example is of this is `OffsetDateTime`.

farazfazli:
@LouisGariepy Please let me know your thoughts on this whenever you get a chance.

LouisGariepy:
Hi Faraz! Sorry for the delay.

I do believe it would be possible to implement Deserialize for the "owned" variant of generated types.

I'm not sure about this comment though

> I believe the struct fields that aren't being deserialized will have to be wrapped with Option, to ensure that deserialization is performed properly. We can't use #[serde(skip_deserializing)] because certain types which lack a default implementation are defined in a different package.

Could you clarify what you mean by "fields that aren't being deserialized"? Do you mean that you would want you api to accept missing fields?

In the structs generated by Cornucopia, Option means that the attribute is nullable in the database. We can't make fields optional based on whether they are `Default` or not.

In my opinion, the `Serialize` and `Deserialize` implementations are provided as conveniences for simple cases. If you need more complex Serialization/Deserialization (for example, making some fields skippable during deserialization), you should consider:
1. Making a deserialization wrapper that handles the deserialization logic, or
2. Marking the relevant types as nullable in your database.

@farazfazli Maybe I'm misunderstood what you meant though. Please let me know and don't hesitate to correct me if that's the case.

Thanks for your contributions!

occuros:
Is this PR still considered?

LouisGariepy:
This should now be supported via [type mappings](https://cornucopia-rs.github.io/cornucopia/configuration.html#custom-type-mappings) and [configured derives](https://cornucopia-rs.github.io/cornucopia/configuration.html#derive-traits).

Issue #204, opened 2023-03-22
Use different type to represent DECIMAL than rust_decimals::Decimal


xoac:
Hi cornucopia team, 
I have found another place where it's hard to work with cornucopia,. It's completely fine to insert `7777777777777777777777777777777777777777777777777777` into PostgreSQL DECIMAL column.

Unfortunately `rust_decimal::Decimal::from_str('7777777777777777777777777777777777777777777777777777')` will return error so using it to represent DECIMAL is limited. 

On possible solution:
1. add PR for bigdecimal crate to optionally enable support for postgresql?
2. migrate from rust_decimal to bigdecimal for DECIMAL type

LouisGariepy:
Looking a bit more into this, it seems like `bigdecimal` has a number of concerning implementation issues.

The maintainer of `rust_decimal` has expressed interest in implementing arbitrary precision https://github.com/paupino/rust-decimal/issues/562#issuecomment-1381169106. This might be something we just have to wait for unless we can find another decent big decimal crate.

LouisGariepy:
See also this comment on why Borsch (a serialization format) didn't implement `bigdecimal` support https://github.com/near/borsh-rs/pull/91#issuecomment-1277935244.

LouisGariepy:
The default still uses rust-decimal, but this is something that can now be configured via [custom type mappings](https://cornucopia-rs.github.io/cornucopia/configuration.html#custom-type-mappings)

PR #206, opened 2023-03-23
Switch rust_decimal to BigDecimal, and increase precision to match Postgres NUMERIC type precision

farazfazli:
Addresses #204, WIP. Currently, facing an error due to BigDecimal not implementing the FromSql<'a> trait.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,17 +29,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ahash"
-version = "0.8.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
-dependencies = [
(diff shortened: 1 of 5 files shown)


farazfazli:
@xoac @LouisGariepy Any ideas on how to fix this? Looks like due to language design, I can't add an additional trait to an external type.

LouisGariepy:
@farazfazli That's right. To implement a trait, you must either own the type or the trait. This is called "orphan rules".

farazfazli:
@LouisGariepy Makes sense, thanks for explaining. I will look into what's needed to open a PR upstream.

farazfazli:
Looks like the BigDecimal crate is hardly maintained, and rust_decimal has a hard limit of 28 digits for precision: https://github.com/paupino/rust-decimal/issues/562

farazfazli:
Closing this PR for now, we can create a new one if a crate is found which satisfies our requirements.

PR #208, opened 2023-03-29
feat: Support custom PostgreSQL Docker images

jacobsvante:
The main use case is to support PostgreSQL extensions, in my case [TimescaleDB](https://hub.docker.com/r/timescale/timescaledb).

This PR also includes improvements to the container healthcheck and supports changing port and container name.

This is a breaking change because some public functions have had their signatures changed.

### DIFF
--- a/crates/cornucopia/src/container.rs
+++ b/crates/cornucopia/src/container.rs
@@ -2,44 +2,73 @@ use std::process::{Command, Stdio};
 
 use self::error::Error;
 
+static DEFAULT_CONTAINER_NAME: &str = "cornucopia_postgres";
+static DEFAULT_DOCKER_IMAGE: &str = "docker.io/library/postgres:latest";
+static DEFAULT_PORT: u16 = 5435;
+
+/// Container related options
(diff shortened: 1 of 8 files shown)


crates/cornucopia/src/container.rs
```diff
@@ -2,58 +2,93 @@ use std::process::{Command, Stdio};
...
     let slow_threshold = 10 + max_retries / 10;
     let mut nb_retries = 0;
-    while !is_postgres_healthy(podman)? {
+    let mut consecutive_successes = 0;
+    while consecutive_successes < 5 {
+        if is_postgres_healthy(opts)? {
+            consecutive_successes += 1;
+        } else {
+            consecutive_successes = 0;
+        }
```
heksesang, line 87: Checking the container logs to see if the "PostgreSQL init process complete; ready for start up." line has been printed is probably more robust than relying on an arbitrary amount of successes.

LouisGariepy:
Now supported after the Clorinde merge. Images can be configured via `container-image`.

PR #210, opened 2023-04-01
Use std SystemTime for timestamps

Virgiel:
Although not as convenient as `time` or `chrono`, the use of `SystemTime` does not require the use of any library.


LouisGariepy:
Can now be configured via [type mappings](https://cornucopia-rs.github.io/cornucopia/configuration.html#custom-type-mappings)

Issue #209, opened 2023-03-31
Feature request: Ability to choose chrono instead of time


jacobsvante:
Currently time fields seem bound to the `time` library. Working with time zones with `time` is a PITA so I was saddened to see that cornucopia does not support chrono.

Might implement this myself after Easter. @LouisGariepy @Virgiel Should I implement as a crate feature, e.g. `chrono` + `time` where `time` is default?

Or maybe #207 would solve this?

Virgiel:
The simplest solution for now, would be to use `SystemTime` for `timestamp` and `timestamptz` and let the user match to `time` or `chrono`. Latter we could add a feature to make this more ergonomic.

jacobsvante:
I didn't know SystemTime can be used to store historical dates..?

Either way I think the feature approach I mentioned might make more sense though, as postgres-types does it this way. I don't think it makes use of SystemTime at all.

LouisGariepy:
@jacobsvante I didn't add chrono support originally because of the whole unsoundness debacle. I tried to not include crates with serious known vulnerabilities.

This has been fixed in version 4.20, but only if you disable default features and don't enable the `oldtime` feature. The sound implementation will be the default in a future 0.5.0 release https://github.com/chronotope/chrono/issues/970.

In any case, disregarding the soundness issues, this will probably be actionable via the new codegen architecture #211.

jacobsvante:
Okay, sounds good 👍

LouisGariepy:
Chrono is now the default, time can still be used via [custom type mappings](https://cornucopia-rs.github.io/cornucopia/configuration.html#custom-type-mappings). See [the migration to 1.0 guide](https://cornucopia-rs.github.io/cornucopia/introduction/migration_to_1_0.html#chrono-instead-of-time)

Issue #186, opened 2022-12-01
A new codegen architecture.


LouisGariepy:
It was brought to my attention over a discussion on Discord that we could improve our current architecture by generating a *crate* instead of a single file for our codegen.

The main benefit of this is that it would allow us to automatically generate a `Cargo.toml` file customized to support all the necessary dependencies and features required by the user's queries, *without polluting their own manifest*. This would hide all the complexity of the current setup while ensuring automatically that compatible versions and features are used. This is all done without tampering with the user's own `Cargo.toml` so there are less risks of unwanted side effects. This would be a neat solution to #163. We also wouldn't have to read and update the toml files, we could simply generate the whole manifest, which is much simpler.

Another benefit is that it would allow us more freedom to organize the generated code in a readable/understandable way instead of having to cram everything into a single file (e.g. db types, sync/async #176 ). Taking this further, we could even generate the client crates directly into this generated crate which would save us from having to publish the 3 client crates.

As a side-effect benefit, cargo can parallelize work over crates, so if you have a lot of queries/generated code, this will improve your compile times.

So, in summary,
* _Much_ better user experience
* Better compile times
* More readable/scalable codegen.
* Gives us more flexibility over generated code and its dependencies in general (future-proof).

# Drawback
Cargo won't publish crates with path dependencies (yet, and the maintainers seem to think its too large of a change to ever be merged), so the user either has to
1) Not publish their crate.
2) Publish the generated crate separately.
This is true for any internal crate so I don't think its such a big drawback, but some users might find that annoying nonetheless.

We could keep the current workflow under a "manual configuration" option (where the user must manually declare the generated modules and manage the correct dependencies) for cases where this is absolutely unwanted.

jacobsvante:
I think this sounds like an great idea. Looking forward to a more readable code layout.

> 1. Not publish their crate.

No other crates are currently depending on any of the cornucopia crates according to crates.io. So I wonder how much of a drawback this really is. I don't see what the use-case would be for depending on it in a library.

Virgiel:
I see all the benefits of using this architecture, but we must consider that this is a significant architectural change for the Rust ecosystem. We will be, to my knowledge, the first code-generation tool to do this. 

I also wonder how we could let the user choose the dependencies version if we generate the `Cargo.toml` file.

dlight:
I think [perseus](https://github.com/framesurge/perseus) also does this (or used to do, not sure)

LouisGariepy:
[...]

PR #211, opened 2023-05-08
New codegen architecture

Virgiel:
A new codegen architecture generating a whole crate instead of a single file #186.

## Changes

- Generate a full crate instead of a file
- Inline clients code
- Minimum amount of dependencies
- Most features are now automatically enabled, only `deadpool` remain
- Split the gigantic generated code file into modules
- No longer generate a named module for postgres `public` implicit schema
- Improve codegen

## Current problems

- Cargo cannot have multiple package with the same name and doesn't support
  nested package. Therefore, I had to use a different name for each generated
  crate.
- We generate way more code than before, and codegen diff are getting out of control

## Remaining improvements

- Test for types in a schema
- Update documentation, examples, website, etc

## Deferred to future PR

- Use [`cargo-px`](https://github.com/LukeMathWalker/cargo-px) instead of `build.rs` for the `auto-build` example. Recommend this it by default instead of the cli ?
- We have the opportunity to change how we handle certain types, in particular times #209 and decimals #204. We could provide options, for example for times, you could choose between : **std** for `SystemTime`, **time** or **chrono**.


Virgiel:
@LouisGariepy Although it is not completely ready for a merge, the core logic is already there and works. This is already a huge PR in itself, so I deferred some of the work for later. I didn't add any new features or made any logical changes to the generated code.

LouisGariepy:
@Virgiel Thanks for the awesome work! I'm reviewing the code as I'm writing this. Might take a bit since there's a lot to look at, but what I have seen so far is promising.

LouisGariepy:
Oh I just realized this was still a draft :sweat_smile:

Virgiel:
I still need to add some error handling, but the new architecture is ready for review !

Virgiel:
I've been using this PR in productions for a few weeks and it works very well !

Virgiel:
A final pain point is that we generate too much code and it makes the PRs hard to read, I suggest putting the examples and reference codegen in gitignore and relying on `cargo px` to generate them on demand. We will continue to check that the codegen works and that the examples run.

PR #214, opened 2023-05-20
Changed generate_managed to generate_against

### LINKED ISSUE #213: CLI command names and consistency.
In our CLI, we renamed the `managed` command to `schema` at some point. But we didn't change the `generate_managed` function to reflect that.

One thing to note is that `generate_schema` sounds a bit misleading as we are generating against a schema, not generating the schema itself.

Should we leave things as is and live with the discrepancy? Should we find a better name to make the CLI and the api consistent?

terracottabucket:
Resolves #213
I though this might be a good substitution, while it is not consistent with the usage of `schema` it could make the command a bit more intuitive.

### DIFF
--- a/crates/cornucopia/src/cli.rs
+++ b/crates/cornucopia/src/cli.rs
@@ -2,7 +2,7 @@ use std::path::PathBuf;
 
 use clap::{Parser, Subcommand};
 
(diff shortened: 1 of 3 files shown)


LouisGariepy:
The issue is still present, but since the codebase has changed so much (and so much time has passed...) let's close this PR for now. If you ever want to pick it up again let me know. Thank you for your contribution.

PR #215, opened 2023-05-24
Optimising codegen using an asynchronous client and pipelining

Virgiel:
When benchmarking codegen, I found that we spent most of our time communicating with the database synchronously to prepare each query one by one. By using an asynchronous client and relying on pipelining, codegen's time went from 22ms to 6ms on my machine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -442,12 +442,14 @@ dependencies = [
  "chumsky",
  "clap",
  "codegen_template",
+ "futures",
  "heck",
  "indexmap",
  "miette",
- "postgres",
  "postgres-types",
  "thiserror",
+ "tokio",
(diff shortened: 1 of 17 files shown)


Virgiel:
When we generate code against a managed database, most of our time is spent on docker commands, 1.3s when managed vs 50ms when live. We don't have an easy way to improve this.

Parsing is also an expensive CPU task that we can parallelize by module, but the gain is less: 6ms down to 4ms. Finally, we could also use mmap instead of reading queries and schema files into memory for even smaller but easy gain.

LouisGariepy:
Since the Clorinde merge, codegen query prepares are now done asynchronously! Thank you for the contribution.

PR #245, opened 2024-04-24
Fix container healthcheck

### LINKED ISSUE #244: Auto build don't work
Running `cargo build` using the `auto_build` example produces one of 2 errors:

```
  --- stderr
  Error: Connection(Error(Error { kind: Closed, cause: None }))
```

Or:

```
  --- stderr
  Error: Connection(Error(Error { kind: Io, cause: Some(Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" }) }))
```

I suspect it is related to how the docker container is brought up? Maybe there's a timing issue here because the container isn't yet ready?

feelingsonice:
Previous container creation logic had the following 2 issues afaik:

1. Health check was too quick. Increasing this from `50` to `500` fixed the problem for me on M2 Max.
2. Container setup logic didn't do cleanup after failure. Resulted in error messages like `Conflict. The container name \"/cornucopia_postgres\" is already in use by container` on subsequent startup.

Misc:

- Fix the build script for `auto_build` example.

This fixes https://github.com/cornucopia-rs/cornucopia/issues/244

### DIFF
--- a/crates/cornucopia/src/container.rs
+++ b/crates/cornucopia/src/container.rs
@@ -5,7 +5,7 @@ use self::error::Error;
 /// Starts Cornucopia's database container and wait until it reports healthy.
 pub fn setup(podman: bool) -> Result<(), Error> {
     spawn_container(podman)?;
-    healthcheck(podman, 120, 50)?;
+    healthcheck(podman, 120, 500)?;
     Ok(())
 }
 
--- a/examples/auto_build/src/cornucopia.rs
+++ b/examples/auto_build/src/cornucopia.rs
@@ -1,93 +1,62 @@
 // This file was generated with `cornucopia`. Do not modify.
 
-#[allow(clippy::all, clippy::pedantic)]
-#[allow(unused_variables)]
-#[allow(unused_imports)]
-#[allow(dead_code)]
-pub mod types {}
-#[allow(clippy::all, clippy::pedantic)]
-#[allow(unused_variables)]
-#[allow(unused_imports)]
-#[allow(dead_code)]
-pub mod queries {
-    pub mod module_1 {
-        use cornucopia_async::GenericClient;
-        use futures;
-        use futures::{StreamExt, TryStreamExt};
-        pub struct StringQuery<'a, C: GenericClient, T, const N: usize> {
-            client: &'a C,
-            params: [&'a (dyn postgres_types::ToSql + Sync); N],
-            stmt: &'a mut cornucopia_async::private::Stmt,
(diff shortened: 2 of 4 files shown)


crates/cornucopia/src/container.rs
```diff
@@ -5,7 +5,7 @@ use self::error::Error;
 /// Starts Cornucopia's database container and wait until it reports healthy.
 pub fn setup(podman: bool) -> Result<(), Error> {
     spawn_container(podman)?;
-    healthcheck(podman, 120, 50)?;
```
Gerald-Hu, line 8: Why too quick is a trouble? And do we really want to 10x it?

examples/auto_build/src/cornucopia.rs
Gerald-Hu, line 1: Can barely understand this file and I hate lifecycle lol... Is this a generated file?

crates/cornucopia/src/container.rs
feelingsonice: This health check is pinging the Postgres container as it starts up, because the container needs sometime to start, ping it too quickly means the container won't be ready yet.

examples/auto_build/src/cornucopia.rs
feelingsonice: Its auto generated.

crates/cornucopia/src/container.rs
heksesang: But it retries the pings, why would it be a problem that the container isn't ready yet?

feelingsonice: I think the health checker is erroring out before it could do any retry:

```rust
fn healthcheck(podman: bool, max_retries: u64, ms_per_retry: u64) -> Result<(), Error> {
    let slow_threshold = 10 + max_retries / 10;
    let mut nb_retries = 0;
    while !is_postgres_healthy(podman)? { // <------ this won't execute for me
       
    }
    // ...
}
```

I haven't confirmed this but a quick Look suggests that `is_postgres_healthy` is calling a command on the `cornucopia_postgres` container before it's even created:


```rust
/// Checks if Cornucopia's container reports healthy
fn is_postgres_healthy(podman: bool) -> Result<bool, Error> {
    Ok(cmd(
        podman,
        &["exec", "cornucopia_postgres", "pg_isready"],
        "check container health",
    )
    .is_ok())
}
```

heksesang: But `is_postgres_healthy` can only return an `Ok(true)` or `Ok(false)`, so it cannot actually error out?

Issue #250, opened 2024-06-13
Support deadpool-postgres 0.14


jacobsvante:
As the title suggests.

LouisGariepy:
Dependencies have been updated! Thank you.

PR #252, opened 2024-06-13
CI: Nightly job to upgrade compatible dependencies

jacobsvante:
Merging this PR will enable a new Github workflow that runs [cargo-edit's upgrade command](https://github.com/killercup/cargo-edit?tab=readme-ov-file#cargo-upgrade) every night. Then tests are run and finally a PR is created if any changes are noticed.

IMHO next step is to automate the release process 😀

Please note that your unit tests are currently failing with a lifetime error because of an old subdependency of yours called rustc-serialize. Merging #250 will fix that problem.

Also note that this PR updates the existing CI workflow to get rid of deprecation warnings.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,47 +7,25 @@ jobs:
     name: Test Suite
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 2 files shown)


LouisGariepy:
Not planned for now, Dependabot does the job fine at the moment. I'm open to revisit this in the future if needed :) Thank you for your contribution.

PR #255, opened 2024-07-17
Add support for type mapping

### LINKED ISSUE #246: Support for type mapping
Similar to Kanel's [customTypeMap](https://kristiandupont.github.io/kanel/configuring.html#customtypemap). This would allow treating PG extension types as some other types that they behave as (e.g., map `public.ulid` to `String`).

duarten:
Similar to Kanel's customTypeMap.

One use case for this is allowing one to treat PG extension types as some other types that they behave as (e.g., map public.ulid to String).

This allows treating a type as some custom type, requiring the user to implement `ToSql`, `FromSql`, and a `XxxBorrowed` variant on it. Additionally, we treat mapping to `String` as a special case.

This is a POC for #246. Happy to improve the PR and add tests if the direction sounds good.

### DIFF
--- a/benches/codegen.rs
+++ b/benches/codegen.rs
@@ -17,6 +17,7 @@ fn bench(c: &mut Criterion) {
                     gen_sync: true,
                     gen_async: false,
                     derive_ser: true,
+                    config: Default::default(),
                 },
             )
             .unwrap()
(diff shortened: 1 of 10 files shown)


LouisGariepy:
Type mapping has been added as part of the Clorinde merge. Thank you for opening this PR.

PR #269, opened 2026-05-05
Disable default features in tokio-postgres

Destructor17:
tokio-postgres is able to run on Cloudflare workers. Cloudflare workers use wasm32-unknown-unknown arch which is not supported by Tokio. Default features in tokio-postgres mean dependency on Tokio. It would be nice to use Cornucopia in Cloudflare workers.


Destructor17:
Ow... just noticed that the last commit was 3y ago. Could you please add a note to README about Cornucopia status, suggest Clorinde as an alternative, and archive this repo? I spend the whole day migrating from ORM to Cornucopia just to figure out it is dead, lol

PR #304, opened 2026-05-20
fix: support empty postgres enum variants

### LINKED ISSUE #289: Postgres enum types with an empty string as a valid value will break all parsing
> Migrated from [halcyonnouveau/clorinde#92](https://github.com/halcyonnouveau/clorinde/issues/92), originally reported by @KenAKAFrosty on 2025-04-16.

### Not sure if this should even be considered a bug. 
Possibly something to just mention in the docs or a FAQ. 

If you have an enum type like  `CREATE TYPE badge AS ENUM ('staff', 'maintenance', 'guest', '')` , where you have an empty string value as a valid  part of the enum, expect to see this error when running the CLI if your queries return that column:
```
thread 'main' panicked at ...
Ident is not allowed to be empty; use Option<Ident>
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Makes sense when you consider that the created  Rust enum can't (and shouldn't) support an empty value inherently

> One could argue that the type shouldn't contain the empty value, but instead allow for nullable columns where that type is used.  I'm inclined to agree, but I'm short on a time machine to convince my old self of that 😅

Not a huge deal as I can work around this now that I know the issue. But if someone was dipping their toes into clorinde, and hit this kind of error, it may put them off. Would be a shame since they wouldn't get to see how awesome clorinde is!

Happy to hear thoughts on this. And I'm open to help with getting it somewhere in the docs if that's something we want!

---

## Original discussion

**@beanpuppy** ([2025-04-17](https://github.com/halcyonnouveau/clorinde/issues/92#issuecomment-2812187022)):
> Thanks for bringing this up. Having an empty string value in an enum doesn't seem like a very good idea, so I wouldn't really want to add support for this even if we could.
>
> I'm not sure if we need to document it anywhere as I doubt many people would do something like this. But since it is technically valid postgres to do this, and we aim to have complete postgres support for every possible thing it could do, I can see how we might want it documented somewhere just for completeness and posterity.
>
> We don't have a FAQ, but I'm not opposed to the idea of it, I don't know what else could be added to it though.
>
> We could also have a "known limitations" document for things like this. Although the only other thing I can think of right now to add to it is halcyonnouveau/clorinde#86, but there's probably more if I dug deeper and loosened my criteria for "limitation".
>
> If it were just left to me, I'd probably leave it here and refer to this issue if anyone else comes and asks for it. But if you're *really* wanting to write some docs, you're free to open a PR with whatever you think is best.

wyf027:
## Summary
- generate a valid Rust enum variant when a Postgres enum contains an empty string value
- reuse the same variant-name fallback in both custom type and SQL conversion generation
- add regression tests covering the empty Postgres enum variant case

Fixes #289.

## Test Plan
- `cargo fmt --check`
- `cargo test -p cornucopia -- codegen::types::tests`
- `cargo test -p cornucopia`
- `cargo clippy -p cornucopia -- -D warnings`
- `git diff --check`

### DIFF
--- a/src/codegen/types.rs
+++ b/src/codegen/types.rs
@@ -135,11 +135,7 @@ fn gen_custom_type(
                 .map(|v| {
                     format_ident!(
                         "{}",
-                        if config.style.enum_variant_camel_case {
-                            v.rs.to_upper_camel_case()
-                        } else {
-                            v.rs.clone()
-                        }
+                        enum_variant_ident_name(v, config.style.enum_variant_camel_case)
                     )
                 })
                 .collect();
@@ -360,16 +356,7 @@ fn enum_sql(
 
     let rs_variants: Vec<_> = variants
         .iter()
-        .map(|v| {
-            format_ident!(
-                "{}",
-                if variant_camel_case {
-                    v.rs.to_upper_camel_case()
-                } else {
-                    v.rs.clone()
-                }
-            )
-        })
+        .map(|v| format_ident!("{}", enum_variant_ident_name(v, variant_camel_case)))
         .collect();
 
     let db_variants: Vec<_> = variants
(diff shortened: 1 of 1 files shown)


LouisGariepy:
Thanks for opening this PR. I added some thoughts [here](https://github.com/cornucopia-rs/cornucopia/issues/289#issuecomment-4499420893), in addition to what beanpuppy has already said. Please let me know what you think :)

